PR #3, opened 2016-06-21
testing/feature/atsd 1475

unrealwork:
**[SQL PERIOD interpolation](https://github.com/axibase/atsd-docs/blob/master/api/sql/examples/interpolate.md) test** releated to [#1475](redmine/issues/1475) issue


src/test/java/com/axibase/tsd/api/sql/data/period/interpolation/constant-interpolation.sql
```diff
@@ -0,0 +1,5 @@
+SELECT entity, date_format(PERIOD(5 minute)), COUNT(value)
```
Sild: At first, move all plain text files to resource folder. For example to
src/test/resources/sql-query

src/test/java/com/axibase/tsd/api/method/entity/EntityMethod.java
```diff
@@ -46,15 +46,14 @@ public void createOrUpdateCheck(final Entity entity) throws IOException {
...
-    Boolean entityExist(final Entity entity) throws IOException {
+    public Boolean entityExist(final Entity entity) throws IOException {
         return entityExist(entity, true);
     }
 
     Boolean entityExist(final Entity entity, Boolean allowExtraFields) throws IOException {
         String entityJson = jacksonMapper.writeValueAsString(entity);
 
         AtsdHttpResponse response = httpSender.sendGet(METHOD_ENTITIES + entity.getName());
-        assertEquals(200, response.getCode());
```
Sild: The main goal was to throw an assert if query is wrong and abort test immediate.

In your solution we should check response in client code.

Once more:
we send wrong query => throw an assert, it's an error in 'framework' code.
we did not find required entity => return false, handle it in client code.

src/test/java/com/axibase/tsd/api/sql/period/interpolation/SqlPeriodTest.java
```diff
@@ -0,0 +1,254 @@
...
+        EntityMethod testEntityMethod = new EntityMethod();
+        EntityMethod.prepare();
+        testEntityMethod.createOrUpdateCheck(testEntity);
+        MetricMethod testMetricMethod = new MetricMethod();
+        testMetricMethod.createOrReplaceMetric(testMetric);
+        boolean created;
+        do {
+            created = testEntityMethod.entityExist(testEntity) && testMetricMethod.metricExists(testMetric);
+        }while (!created);
+        Series series = new Series();
```
Sild: Why dont use `new Series(testEntity.getName(), testMetric.getName())` ?
No need to use the default constructor for Series object

src/test/java/com/axibase/tsd/api/sql/period/interpolation/SqlPeriodTest.java
```diff
@@ -0,0 +1,254 @@
...
+    }
+
+    @AfterClass
+    public static void clearTestData() throws Exception {
+        EntityMethod testEntityMethod = new EntityMethod();
+        MetricMethod testMetricMethod = new MetricMethod();
+        testEntityMethod.deleteEntity(TEST_PREFIX + "-entity");
+        new MetricMethod().deleteMetric(TEST_PREFIX + "-metric");
+        boolean deleted;
+        do {
```
Sild: Infinity loop without any timeouts or delays? We hope the metric&entity will be deleted anyway?

src/test/java/com/axibase/tsd/api/sql/period/interpolation/SqlPeriodTest.java
```diff
@@ -0,0 +1,254 @@
...
+        seriesMethod.insertSeries(series);
+        waitForSeriesData(startDate, endDate, samplesCount);
+    }
+
+    @AfterClass
+    public static void clearTestData() throws Exception {
+        EntityMethod testEntityMethod = new EntityMethod();
+        MetricMethod testMetricMethod = new MetricMethod();
+        testEntityMethod.deleteEntity(TEST_PREFIX + "-entity");
+        new MetricMethod().deleteMetric(TEST_PREFIX + "-metric");
```
Sild: Can we use existing object created 2 lines before?

src/test/java/com/axibase/tsd/api/Util.java
```diff
@@ -141,4 +147,66 @@ public static String generateStringFromChar(char c, int length) {
...
         return sb.toString();
     }
+    /**
+     * Parse date from ISO8601 format string
+     *
+     * @param dateString sting in ISO8601 format
+     *                   @see <a href ="https://en.wikipedia.org/wiki/ISO_8601">ISO8601</>
+     * @return Date Date class' object
+     */
+    public static Date parseISODate(String dateString) {
```
Sild: You can use existing method `getDate(String date)` in Util class. But I think we should rename it to be more clear.

src/test/java/com/axibase/tsd/api/sql/SqlExecuteMethod.java
```diff
@@ -0,0 +1,146 @@
...
+    /**
+     * Execute a sql query on ATSD instance,
+     * that described in properties file
+     *
+     * @param sqlQuery query to ATSD, with syntax described in docs.
+     * @return HttpResponse  http responses
+     * @throws IOException for incorrect url.
+     * @see <a href="SQL API">https://github.com/axibase/atsd-docs/tree/master/api/sql</a>
+     */
+    public HttpResponse query(String sqlQuery) throws IOException {
```
Sild: You had to use the following class to execute https queries: 
`com.axibase.tsd.api.transport.http.HTTPClientPure`

But we will replace it to another transport layer soon, so do not hurry to rewrite it now.

unrealwork:
I've fixed last notes, what i should do more?

PR #7, opened 2016-06-22
fix bug with NPE when dev.client.properties is not found

KutuzovGregory:


rmakulov:
Yes, it will be useful, thank you!

oik77:
ok

unrealwork:
:+1:

src/test/java/com/axibase/tsd/api/Config.java
```diff
@@ -50,19 +51,25 @@ public static Logger getLogger() {
...
+            if (tryInitConfig(DEFAULT_CONFIG_FILE)) return instance;
+            throw new FileNotFoundException("*.client.properties not found");
         }
         return instance;
     }
 
+    private static boolean tryInitConfig(String config) {
+        URL configUrl = Config.class.getClassLoader().getResource(config);
+        if (configUrl != null) {
+            logger.debug("Using {} for config", config);
```
Sild: Maybe write "trying to use" instead?

PR #8, opened 2016-06-22
develop/feature/base-method-abstraction

unrealwork:
Created class BaseMethod for non api requests


src/test/java/com/axibase/tsd/base/BaseMethod.java
```diff
@@ -21,26 +20,33 @@
...
 
     static {
         java.util.logging.LogManager.getLogManager().reset();
         SLF4JBridgeHandler.removeHandlersForRootLogger();
         SLF4JBridgeHandler.install();
         java.util.logging.Logger.getLogger("").setLevel(Level.FINEST);
-
+        try {
+            prepare();
+        } catch (FileNotFoundException e) {
```
KutuzovGregory: there is no reason to catch this exception

KutuzovGregory:
Ok

rmakulov:
Ok.

unrealwork:
Merged with master without any conflicts

Sild:
Maybe we should change package tree like the following:

```
builder.\.\.
model.\.\.
transport.tcp
action.api
action.root
```
?

PR #9, opened 2016-06-22
move to common entity\metric\type registry

Sild:


rmakulov:
Ok.

KutuzovGregory:
Ok

unrealwork:
:+1:

src/test/java/com/axibase/tsd/api/Registry.java
```diff
@@ -0,0 +1,30 @@
...
+import java.util.TreeSet;
+
+/**
+ * @author Dmitry Korchagin.
+ */
+public enum Registry {
+    Entity("Entity"), Metric("Metric"), Type("Type");
+
+    private String registryType;
+    private Set<String> registredSet = new TreeSet<>();
```
oik77: If you do not need to support elements order in set, you should use hashset because it has better performance

Sild: Ok, I'll fix it on next iteration
http://korzh.net/2010-11-proizvoditelnost-osnovnyx-kollekcij-java-sravnenie-s-analogami-net.html

PR #16, opened 2016-06-23
add offset field tests

Sild:


src/test/java/com/axibase/tsd/api/method/property/PropertyQueryTest.java
```diff
@@ -1409,6 +1409,138 @@ public void testLimitNegative() throws Exception {
 
     }
```
KutuzovGregory: Why not use the same data for all 3 tests? Just put 3 property and than query them with different offsets. I think that this is the case of parametrised test.

KutuzovGregory:
Why not use the same data for all 3 tests? Just put 3 property and than query them with different offsets. I think that this is the case of parametrised test.

Sild:
Because the tests should be atomic and there is no any reason to use parameterized tests for few similar tests

Sild:
Because the tests should be atomic and there is no any reason to use parameterized tests for few similar tests

KutuzovGregory:
ok

rmakulov:
Ok

PR #39, opened 2016-07-05
API: raise 404 on url not found

rmakulov:


src/test/java/com/axibase/tsd/api/method/series/SeriesInsertTest.java
```diff
@@ -417,4 +420,48 @@ public void testTimeRangeMaxInISOOverflow() throws Exception {
 
         assertEquals(0, seriesList.get(0).getData().size());
     }
+
+    /* #2927 */
+    @Test
+    public void testUrlNotFoundGetRequest() throws Exception {
```
Sild: can you split them to many tests?

KutuzovGregory:
ok

PR #44, opened 2016-07-07
Entities api example

unrealwork:
my entities method example.
What do you think about it?
Usages
```java
Response response = EntitiesMethod
               .entity("atsd")
               .get();
Response response = EntitiesMethod
               .entity("atsd")
               .metrics()
```

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/entity/EntitiesMethod.java
+++ b/src/test/java/com/axibase/tsd/api/method/entity/EntitiesMethod.java
@@ -1,0 +1,109 @@
+package com.axibase.tsd.api.method.entity;
+
+import com.axibase.tsd.api.method.BaseMethod;
+import org.slf4j.Logger;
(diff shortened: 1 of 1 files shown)


rmakulov:
It looks nice. But what about current check methods? Could you please describe architecture? What about sender abstraction?

src/test/java/com/axibase/tsd/api/method/entity/EntitiesMethod.java
```diff
@@ -0,0 +1,109 @@
...
+import org.slf4j.LoggerFactory;
+
+import javax.ws.rs.client.Entity;
+import javax.ws.rs.client.WebTarget;
+import javax.ws.rs.core.Response;
+import java.lang.invoke.MethodHandles;
+import java.util.Map;
+
+/**
+ * @author Dmitry Korchagin.
```
KutuzovGregory: there is no need for this

PR #50, opened 2016-07-11
SQL: timezone format

unrealwork:
## SQL: timezone format
- [x] Added tests for date_format timezone feature

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/date_format/SqlTimezoneFormatTests.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/date_format/SqlTimezoneFormatTests.java
@@ -44,7 +44,7 @@ public class SqlTimezoneFormatTests extends SqlTest {
 
         List<String> expectedColumnValues = Arrays.asList("2016-06-03T09:23:00+0000");
 
-        assertTableColumnValues(expectedColumnValues, resultTable, "f-date");
+        assertTableContainsColumnValues(expectedColumnValues, resultTable, "f-date");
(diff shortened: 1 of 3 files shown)


src/test/java/com/axibase/tsd/api/method/sql/date_format/SqlTimezoneFormatTests.java
```diff
@@ -0,0 +1,101 @@
...
+    /**
+     * issue #2904
+     */
+    @Test
+    public void testSimpleDateFormatWithZeroZone() {
+        String sqlQuery = "SELECT time, date_format(time, \"yyyy-MM-dd'T'HH:mm:ssZ\") AS 'f-date'" +
+                "FROM 'sql-date-format-timezone-metric' " +
+                "WHERE datetime = '2016-06-03T09:23:00.000Z'";
+
+        String expectedFormatDate = "2016-06-03T09:23:00+0000";
```
Sild: extract common logic to common method, for example: https://github.com/axibase/atsd-api-test/blob/master/src/test/java/com/axibase/tsd/api/method/entity/EntityGetPropertyTypesTest.java#L46

unrealwork: It's very specific logic, With certain table  cell, that uses in only that test cases. I don't think that I will reuse it. And one method with big string params will reduce the readability.

oik77:
lgtm

PR #51, opened 2016-07-11
SQL: select all from join table

unrealwork:
## SQL: timezone format
- [x] Added tests for select all from join table

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/select/SqlSelectAllWithJoinTests.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/select/SqlSelectAllWithJoinTests.java
@@ -1,0 +1,125 @@
+package com.axibase.tsd.api.method.sql.select;
+
+import com.axibase.tsd.api.Util;
+import com.axibase.tsd.api.method.sql.SqlTest;
+import com.axibase.tsd.api.model.series.Sample;
+import com.axibase.tsd.api.model.series.Series;
+import com.axibase.tsd.api.model.sql.StringTable;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import java.util.Arrays;
+import java.util.List;
+
+/**
+ * @author Igor Shmagrinskiy
+ */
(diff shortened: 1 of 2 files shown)


src/test/java/com/axibase/tsd/api/method/sql/select/SqlSelectAllWithJoinTests.java
```diff
@@ -0,0 +1,145 @@
...
+import java.util.HashSet;
+import java.util.Set;
+
+import static org.junit.Assert.assertEquals;
+
+/**
+ * @author Igor Shmagrinskiy
+ */
+public class SqlSelectAllWithJoinTests extends SqlTest {
+    private static String TEST_PREFIX = "sql-select-all-join-";
```
oik77: final

src/test/java/com/axibase/tsd/api/method/sql/select/SqlSelectAllWithJoinTests.java
```diff
@@ -0,0 +1,145 @@
...
+                "sql-select-all-join-metric-2.entity",
+                "sql-select-all-join-metric-1.value",
+                "sql-select-all-join-metric-2.value",
+                "sql-select-all-join-metric-1.datetime",
+                "sql-select-all-join-metric-2.datetime",
+                "sql-select-all-join-metric-1.tags.a",
+                "sql-select-all-join-metric-2.tags.a",
+                "sql-select-all-join-metric-2.tags.b"
+        ));
+        Set<String> resultColumnNames = extractColumnNames(executeQuery(sqlQuery)
```
Sild: this part also can be extract to common function

oik77:
I'd better add constants with metric names

unrealwork:
@oik77 I think is not useful for someone who will be read this tests. What we have do when amount of metrics in this test case will grow to few dozen. It will reduce test's readability.

oik77:
@unrealwork if we increase number of metrics, we will need big refactoring

oik77:
LGTM

PR #60, opened 2016-07-18
SQL: Metric tags, additional case sensitivity and `dash-name`  tests(squashed)

unrealwork:
- added tests for dash in tag's name


Sild:
ok

rmakulov:
lgtm

PR #64, opened 2016-07-22
CSVUploadTest: transform expectedDate to server TZ

rmakulov:


src/test/java/com/axibase/tsd/api/method/version/VersionMethod.java
```diff
@@ -0,0 +1,15 @@
+package com.axibase.tsd.api.method.version;
+
+import com.axibase.tsd.api.method.BaseMethod;
+
+import javax.ws.rs.core.Response;
+
+public class VersionMethod extends BaseMethod {
```
Sild: should we leave it here or move to BaseMethod?

rmakulov: Let's move.

Sild:
review done

Sild:
ok

PR #71, opened 2016-07-23
SQL: group by datetime tests 

unrealwork:
- [x] Added group by datetime tests


Sild:
lgtm

rmakulov:
Why only one value for grouping?

unrealwork:
@rmakulov I check only syntax in these tests.

rmakulov:
lgtm

PR #79, opened 2016-07-27
SQL: tags' name column  with double quotes

unrealwork:
## Tags' name column  with double quotes
- [x] tests


rmakulov:
lgtm

KutuzovGregory:
ok

PR #80, opened 2016-07-27
SQL: select entity tags tests

unrealwork:
## SQL: select entity tags tests
- [x] select entity tags tests
- [x] some useful methods replaced to `SqlMethod `


src/test/java/com/axibase/tsd/api/method/sql/where/SqlEntityTagsTest.java
```diff
@@ -32,22 +32,10 @@ public static void prepareDate() {
     }
 
 
-    /**
+    /*
```
Sild: /**

Sild:
lgtm

rmakulov:
lgtm

PR #81, opened 2016-07-27
1278 urlencode tests

Sild:


src/test/java/com/axibase/tsd/api/method/entity/EntityMethod.java
```diff
@@ -56,14 +56,15 @@ public static Response updateEntity(com.axibase.tsd.api.model.entity.Entity enti
     public static Response deleteEntity(String entityName) throws Exception {
         Response response = httpApiResource.path(METHOD_ENTITY).resolveTemplate("entity", entityName).request().delete();
         response.bufferEntity();
+        String json = formatToJsonString(response);
```
rmakulov: Have you forgot to do smth with that string?

src/test/java/com/axibase/tsd/api/method/entityGroup/EntityGroupCreateOrReplaceTest.java
```diff
@@ -0,0 +1,58 @@
...
+    public void testNameContainsCyrillic() throws Exception {
+        EntityGroup entityGroup = new EntityGroup("urlencodecreateReplaceйёentitygroup3");
+        assertUrlEncodePathHandledCorrectly(entityGroup);
+
+    }
+
+    public void assertUrlEncodePathHandledCorrectly(final EntityGroup entityGroup) throws Exception {
+        if (entityGroupExist(entityGroup)) {
+            deleteEntityGroup(entityGroup.getName());
+        }
```
rmakulov: I thought, we suppose that the database is empty - am I wrong?
Why do we delete entity group?

src/test/java/com/axibase/tsd/api/method/entity/EntityMethod.java
Sild: No. It's redundant action, I'll remove it.

src/test/java/com/axibase/tsd/api/method/entityGroup/EntityGroupCreateOrReplaceTest.java
Sild: Yes, I execute it "just in case".
For example to debug reason on local machine over the filled database

src/test/java/com/axibase/tsd/api/method/series/SeriesUrlQueryTest.java
```diff
@@ -0,0 +1,108 @@
...
+    private void assertUrlEncodePathHandledCorrectly(Series series) throws IOException {
+        series.addData(new Sample(Util.MIN_STORABLE_DATE, "0"));
+        insertSeriesCheck(series);
+        Map<String, String> parameters = new HashMap<>();
+        parameters.put("startDate", Util.MIN_QUERYABLE_DATE);
+        parameters.put("endDate", Util.MAX_QUERYABLE_DATE);
+
+        Response response = urlQuerySeries(series.getEntity(), series.getMetric(), parameters);
+        assertEquals(OK.getStatusCode(), response.getStatus());
+        List<Series> responseSeries = response.readEntity(new GenericType<List<Series>>() {
```
unrealwork: What about  list's comparing?Maybe it contains an extra element. Or Series contain extra data element.  You don't check it.  List can contains different series amount. Better make Series comparable and make 
```java
assertEquals(expectedSeries,resonseSeries);
```

Sild: I'm not interesting in strict compare of series here. This test check `url encode` for path parameters, so if I get same `Sample`, `metricName` and `entityName` it's enough.
Also it will enough to check only `entityName` and `metricName`.  I check other fields 'just in case'

rmakulov:
lgtm

KutuzovGregory:
ok

PR #87, opened 2016-08-02
SQL: IS NULL with NaN values tests

unrealwork:
# SQL: IS NULL with NaN values tests (squashed)
- [x] Added tests for IS NULL operator applied to NaN values


Sild:
ok

rmakulov:
lgtm

PR #100, opened 2016-08-12
wildcard in entities + alertshistory

Sild:


src/test/java/com/axibase/tsd/api/transport/tcp/TCPSender.java
```diff
@@ -70,21 +69,21 @@ public void send(String command, long sleepDuration) throws IOException, Interru
...
+    public void sendCheck(String command) throws Exception, InterruptedException {
         setCommand(command);
         boolean successed = sendDebugMode();
         if (!successed)
-            throw new IOException("Fail to check inserted command");
+            throw new Exception("Fail to check inserted command");
     }
 
-    public void sendCheck (String command, long sleepDuration) throws IOException, InterruptedException {
+    public void sendCheck(String command, long sleepDuration) throws Exception, InterruptedException {
```
unrealwork: `InterruptedException` inherits from `Exception`

src/test/java/com/axibase/tsd/api/method/property/PropertyQueryTest.java
```diff
@@ -651,7 +650,7 @@ public void testOnlyTypeSpecified() throws IOException {
      * #NoTicket
      */
     @Test
-    public void testEntityFilterEndDateAbsent() throws IOException {
+    public void testEntityFilterEndDateAbsent() throws Exception {
         Map<String, Object> request = new HashMap<>();
```
unrealwork: Why String, Object?

src/test/java/com/axibase/tsd/api/method/message/MessageQueryTest.java
```diff
@@ -111,6 +127,76 @@ public void testMillisecondsUnsupported() throws Exception {
...
+        insertMessageCheck(message);
+
+        Map<String, Object> query = new HashMap<>();
+        query.put("entities", Arrays.asList("message-query-wildcard-2*"));
+        query.put("startDate", message.getDate());
+        query.put("endDate", Util.addOneMS(message.getDate()));
+
+        final String given = formatToJsonString(executeQuery(query));
+        final String expected = jacksonMapper.writeValueAsString(Arrays.asList(message));
+        assertTrue(compareJsonString(expected, given));
```
unrealwork: Specify assert with suitable  message

src/test/java/com/axibase/tsd/api/method/property/PropertyQueryTest.java
Sild: If I get you correctly, It's not what we need:

```
List<String> test = new ArrayList<>();
test.add("a");
test.add("b");
Map<String, String> testobj = new HashMap<>();
testobj.put("testarray", test.toString());
System.out.println(jacksonMapper.writeValueAsString(testobj)); //{"testarray":"[a, b]"}
```

unrealwork: Why do you not implement `AlertQuery` like [SeriesQuery](https://github.com/axibase/atsd-api-test/blob/master/src/test/java/com/axibase/tsd/api/model/series/SeriesQuery.java) and Alert Model?

Sild: More actual query why do you implement it)
We have started to write a query as a simple map for more flexibility. After a while we start to use QueryObject to simplify query building.

PR #112, opened 2016-08-19
alert\history query entityExpression filter

Sild:


src/test/java/com/axibase/tsd/api/method/alert/AlertMethod.java
```diff
@@ -66,4 +66,17 @@ public static void generateAlertForEntity(final String entityName) throws Except
...
 
+    public static void generateAlertHistoryForEntity(final String entityName) throws Exception {
+        Series series = new Series();
+        series.setEntity(entityName);
+        series.setMetric(Util.RULE_METRIC_NAME);
+        series.addData(new Sample(Util.ISOFormat(new Date()), Util.ALERT_OPEN_VALUE));
+        SeriesMethod.insertSeriesCheck(series);
+
+        series.setData(null);
+        series.addData(new Sample(Util.ISOFormat(new Date()), Util.ALERT_CLOSE_VALUE));
```
unrealwork: Why do you use new Date() instead constant?

rmakulov:
lgtm

src/test/java/com/axibase/tsd/api/method/alert/AlertMethod.java
Sild: ATSD rules will not apply to series which does not satisfy the following condition:
|currentDate - seriesDate| < VALUE
There VALUE is specified in ATSD (and is about 1 minute)

unrealwork:
ok

PR #130, opened 2016-08-30
add interval field test in series query

Sild:


GregoryKutuzov:
ok

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
```diff
@@ -26,6 +29,7 @@
     private static final String sampleDate = "2016-07-01T14:23:20.000Z";
     private static final Series series;
     private static Calendar calendar = Calendar.getInstance();
+    private static final Logger logger = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
```
rmakulov: Why we can't use here explicit SeriesQueryTest.class?

Sild: That's more common solution, we can just take this definition to another class without any changes. And be sure that logs will be written by right class.

rmakulov:
ok

PR #131, opened 2016-09-02
add serieQuery group tests

Sild:

PR #132, opened 2016-09-02
add extend=false\null tests

Sild:


GregoryKutuzov:
same issues as in https://github.com/axibase/atsd-api-test/pull/131

PR #150, opened 2016-09-19
#2993 with code refactoring

rmakulov:
Tests that checks correctness of processing request with the value "UNKNOWN" the Entity of the field in alerts/query and alerts/history-query.

### DIFF
--- a/src/test/java/com/axibase/tsd/api/Util.java
+++ b/src/test/java/com/axibase/tsd/api/Util.java
@@ -4,7 +4,9 @@ import com.axibase.tsd.api.method.BaseMethod;
 import com.axibase.tsd.api.method.version.Version;
 import com.axibase.tsd.api.method.version.VersionMethod;
 import com.fasterxml.jackson.databind.util.ISO8601Utils;
+import org.json.JSONArray;
 import org.json.JSONException;
+import org.json.JSONObject;
 
 import java.text.ParseException;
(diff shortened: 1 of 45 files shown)


src/test/java/com/axibase/tsd/api/Util.java
```diff
@@ -104,4 +106,12 @@ public static StringBuilder appendChar(StringBuilder sb, char c, int count) {
         }
         return sb;
     }
+
+    public static String extractField(int index, String field, JSONArray array) throws JSONException {
```
unrealwork: more verbose

src/test/java/com/axibase/tsd/api/method/series/SeriesCommandTest.java
```diff
@@ -85,7 +85,7 @@ public void testMaxLengthOverflow() throws Exception {
         JSONArray storedSeriesList = executeQuery(seriesQueries);
 
         for (int i = 0; i < METRICS_COUNT; i++) {
-            Assert.assertEquals("[]", getField(i, "data", storedSeriesList), "Managed to insert command that length is max + 1");
+            Assert.assertEquals("[]", Util.extractField(i, "data", storedSeriesList), "Managed to insert command that length is max + 1");
```
unrealwork: Why  do we not implement  `SeriesQueryResult` and deserialize it from json?

PR #168, opened 2016-10-05
#3371

rmakulov:
Test for wildcard in entity name when exactMatch=true.
ExactMatch can exclude series with tags not specified in the request.

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
@@ -1,7 +1,5 @@
 package com.axibase.tsd.api.method.series;
 
-import com.axibase.tsd.api.Registry;
-import com.axibase.tsd.api.Util;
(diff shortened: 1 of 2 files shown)


src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
```diff
@@ -513,6 +509,35 @@ public void testAggregateNoPeriodRaiseError() throws Exception {
...
+    */
+    @Test
+    public void testEntityWithWildcardExactMatchTrue() throws Exception {
+        String entityNameBase = "series-query-limit-entity-";
+        String metricName = "series-query-limit-metric";
+
+        Series series = new Series(entityNameBase.concat("1"), metricName);
+        series.addData(new Sample(sampleDate, "7"));
+        insertSeriesCheck(Collections.singletonList(series));
+
```
unrealwork: Please, register new entity

PR #180, opened 2016-10-24
SQL: new metric fields #3474

unrealwork:
- [x] Test support of metric fields


GregoryKutuzov:
@unrealwork resolve conflicts

PR #184, opened 2016-11-09
Added tests of `enabled` field (#3550)

VeselovAlex:


VeselovAlex:
Replace with pull request #185

PR #188, opened 2016-11-11
Series data text (x) tests added

VeselovAlex:


src/test/java/com/axibase/tsd/api/method/series/SeriesInsertTest.java
```diff
@@ -688,4 +689,58 @@ public void testTagValueNullRaiseError() throws Exception {
...
+    public void testXTextField(int testN, Object x) throws Exception {
+        String entityName = "e-text-"+testN;
+
+        String metricName = "m-text-"+testN;
+        String largeNumber = "10.1";
+        final long t = 1465485524888L;
+
+        Series series = new Series(entityName, metricName);
+        Sample sample = new Sample(t, largeNumber);
+        sample.setX(x);
```
rmakulov: What if we add constructor from 3 args for Sample and use it in series.addData?

src/test/java/com/axibase/tsd/api/method/series/SeriesInsertTest.java
```diff
@@ -688,4 +689,58 @@ public void testTagValueNullRaiseError() throws Exception {
...
+        series.addData(sample);
+        Metric metric = new Metric();
+        metric.setName(metricName);
+        metric.setDataType(DataType.FLOAT);
```
rmakulov: ATSD by default create metric with float datatype, so here is enough to insert series and metric will be created too. Also as we have added insertSeriesCheck before please use it if it is possible.

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
```diff
@@ -509,6 +510,58 @@ public void testAggregateNoPeriodRaiseError() throws Exception {
...
+    @Test(dataProvider = "dataTextProvider")
+    public void testXTextField(int testN, Object x) throws Exception {
+        String entityName = "e-text-insert-"+testN;
+        String metricName = "m-text-insert-"+testN;
+
+//        Metric metric = new Metric();
+//        metric.setName(metricName);
+//        metric.setDataType(DataType.FLOAT);
+//        MetricMethod.createOrReplaceMetricCheck(metric);
+
```
rmakulov: Please, delete commented code.

rodionos:
1) Test empty value insert. x:"" - OK
2) Test null value. x: - ERROR
3) Test that last text value prevails: insert x:1, insert x:2, read: 2
4) Test that insert of x: field with series tags doesn't override x: for series without tags.
5) Test x: is inserted and its changes are recorded for metric with versioning:true

src/test/java/com/axibase/tsd/api/method/series/SeriesInsertTest.java
```diff
@@ -663,6 +665,167 @@ public void testTagValueNullRaiseError() throws Exception {
...
+        Sample sample = new Sample("2016-10-11T13:00:00.000Z", new BigDecimal(1.0), xText);
+        series.addData(sample);
+        insertSeriesCheck(Collections.singletonList(series));
+
+        series.addTag("foo", "foo");
+        SeriesQuery seriesQuery = new SeriesQuery(series);
+        insertSeriesCheck(Collections.singletonList(series));
+
+        List<Series> seriesList = executeQueryReturnSeries(seriesQuery);
+        assertEquals("Only one series should be responded", 1, seriesList.size());
```
unrealwork: May be, you just try to check equivalent of series' Lists ? 



```java
/*assertEquals("Only one series should be responded", 1, seriesList.size());
        List<Sample> respondedData = seriesList.get(0).getData();
        assertEquals("Only one sample should be responded", 1, respondedData.size());
        assertEquals("Stored text value incorrect", data[data.length - 1], respondedData.get(0).getText());*/
        assertEquals(seriesList, Collections.singletonList(series));
```

PR #189, opened 2016-11-11
SQL: string functions result tests #2920

unrealwork:
- [x] Added test for functionality
- [x] Connection timeout increased to 1 minute

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/function/aggregate/AverageFunction.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/function/aggregate/AverageFunction.java
@@ -1,0 +1,41 @@
+package com.axibase.tsd.api.method.sql.function.aggregate;
+
+
+import com.axibase.tsd.api.method.series.SeriesMethod;
+import com.axibase.tsd.api.method.sql.SqlTest;
+import com.axibase.tsd.api.model.series.Series;
+import com.axibase.tsd.api.util.Mocks;
+import org.testng.annotations.DataProvider;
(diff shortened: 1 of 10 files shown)


src/test/java/com/axibase/tsd/api/model/sql/operator/StringOperators.java
```diff
@@ -0,0 +1,33 @@
+package com.axibase.tsd.api.model.sql.operator;
+
+
+public class StringOperators {
```
GregoryKutuzov: What is the purpose of this class?

unrealwork:
PR is synced with actual master version

src/test/java/com/axibase/tsd/api/method/sql/function/aggregate/AverageFunction.java
```diff
@@ -0,0 +1,29 @@
...
+import com.axibase.tsd.api.util.Mocks;
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+
+
+public class AverageFunction extends SqlTest {
+
+    @Test(dataProvider = "sqlStringArg", dataProviderClass = SqlDataProvider.class)
+    public void testLengthFunction(String param) throws Exception {
```
GregoryKutuzov: It seems that you are using this data provider only in this test. Maybe it would be better to move it to here?

unrealwork: I've replaced it.

PR #190, opened 2016-11-15
SQL: ISNULL empty string handling 

unrealwork:
- [x] Increased connection timeout
- [x] added tests for `ISNULL` function
- [x] Addded [`Mocks`](https://github.com/axibase/atsd-api-test/blob/shmagrinskiy_3557/src/test/java/com/axibase/tsd/api/Util.java#L145) subclass to [`Util`](https://github.com/axibase/atsd-api-test/blob/shmagrinskiy_3557/src/test/java/com/axibase/tsd/api/Util.java) class


src/test/java/com/axibase/tsd/api/model/sql/SqlDataProvider.java
```diff
@@ -0,0 +1,32 @@
+package com.axibase.tsd.api.model.sql;
+
+import org.testng.annotations.DataProvider;
+
+public class SqlDataProvider {
```
GregoryKutuzov: where this class is used?

PR #191, opened 2016-11-15
Tests of entityExpression query added

VeselovAlex:


VeselovAlex:
Overwriiten with #199

PR #213, opened 2016-12-26
SQL: slow summation query #3711

RawrLion:
tests for #3711

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlParseTimeCheck.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlParseTimeCheck.java
@@ -1,0 +1,50 @@
+package com.axibase.tsd.api.method.sql.syntax;
+
+import com.axibase.tsd.api.method.series.SeriesMethod;
+import com.axibase.tsd.api.method.sql.SqlTest;
+import com.axibase.tsd.api.model.series.Sample;
+import com.axibase.tsd.api.model.series.Series;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.util.Arrays;
+
+import static com.axibase.tsd.api.util.Util.TestNames.entity;
+import static com.axibase.tsd.api.util.Util.TestNames.metric;
+
+public class SqlParseTimeCheck extends SqlTest {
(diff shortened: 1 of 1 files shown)


src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlParseTimeCheck.java
```diff
@@ -0,0 +1,50 @@
...
+        SeriesMethod.insertSeriesCheck(Arrays.asList(series1));
+    }
+
+    /**
+     * #3711
+     */
+    @Test(timeOut = 10000)
+    public void testLongParseTime() {
+        String sqlQuery = String.format(
+                "SELECT value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value+value FROM '%s'",
```
GregoryKutuzov: replace with function which repeat value+value N times

src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlParseTimeCheck.java
```diff
@@ -0,0 +1,50 @@
...
+                TEST_METRIC_NAME
+        );
+
+        String[][] expectedRows = {
+                {"27"}
+        };
+
+        assertSqlQueryRows(sqlQuery, expectedRows, "Join without Group by gives wrong result");
```
GregoryKutuzov: Copy pasted irrelevant error message. Fix it.

src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlParseTimeCheck.java
```diff
@@ -0,0 +1,50 @@
...
+        );
+
+        SeriesMethod.insertSeriesCheck(Arrays.asList(series1));
+    }
+
+    /**
+     * #3711
+     */
+    @Test(timeOut = 10000)
+    public void testLongParseTime() {
```
GregoryKutuzov: we actually are testing that this sql request's processing time is sufficient, isn't it? if so please rename the method name.

GregoryKutuzov:
@unrealwork is it ok?

PR #215, opened 2016-12-26
SQL: Math function ABS does not accept 2nd aggregate expression #3738

RawrLion:
added tests for #3738

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/function/math/SqlFunctionMathAbs.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/function/math/SqlFunctionMathAbs.java
@@ -39,11 +39,11 @@ public class SqlFunctionMathAbs extends SqlTest {
     public Object[][] provideTestsDataForAbsTest() {
         return new Object[][]{
                 {
-                        "abs(avg(value))",
+                        "avg(value)",
                         "2"
                 },
                 {
-                        "abs(max(value))",
+                        "max(value)",
                         "3"
                 },
                 {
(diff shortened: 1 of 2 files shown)


src/test/java/com/axibase/tsd/api/method/sql/function/math/SqlFunctionMathAbs.java
```diff
@@ -0,0 +1,86 @@
...
+        );
+
+        SeriesMethod.insertSeriesCheck(Collections.singletonList(series1));
+    }
+
+    @DataProvider
+    public Object[][] provideTestsDataForAbsTest() {
+        return new Object[][]{
+                {
+                        "avg(value)",
```
GregoryKutuzov: why do you add this case?

src/test/java/com/axibase/tsd/api/method/sql/function/math/SqlFunctionMathAbs.java
```diff
@@ -0,0 +1,86 @@
...
+        };
+    }
+
+    /**
+     * #3738
+     */
+    @Test(dataProvider = "provideTestsDataForAbsTest")
+    public void testAbsWithAggregateExpressionsInside(String query, String value) {
+        String sqlQuery = String.format(
+                "SELECT %s FROM '%s'",
```
GregoryKutuzov: if you will remove first case you should replace `"SELECT %s FROM '%s'"` with `"SELECT abs( %s ) FROM '%s'"`

GregoryKutuzov:
also rename com.axibase.tsd.api.method.sql.syntax.SqlParseTimeCheck to com.axibase.tsd.api.method.sql.syntax.SqlParseTimeTest

PR #218, opened 2017-01-04
SQL: LOCATE function raises NPE in WHERE clause #3749

RawrLion:
tests for #3749


src/test/java/com/axibase/tsd/api/method/sql/function/string/SqlFunctionStringLocate.java
```diff
@@ -0,0 +1,118 @@
...
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.Arrays;
+
+import static com.axibase.tsd.api.util.Util.TestNames.entity;
+import static com.axibase.tsd.api.util.Util.TestNames.metric;
+
+public class SqlFunctionStringLocate extends SqlTest {
```
GregoryKutuzov: move to LocateTest

RawrLion: moved

PR #223, opened 2017-01-16
Contribute guide: Pre-Requirements chapter

unrealwork:


src/test/resources/client.properties
```diff
@@ -1,8 +1,10 @@
+# ATSD  instance properties
 login=username
 password=password
 protocol=http
 serverName=hbs.axibase.com
```
rmakulov: Please, update serverName

PR #227, opened 2017-01-19
SQL: CONCAT for numbers #3768

RawrLion:
added tests for #3768


src/test/java/com/axibase/tsd/api/method/sql/function/string/ConcatTest.java
```diff
@@ -1,21 +1,61 @@
...
+
+    private static final String TEST_METRIC1 = metric();
+    private static final String TEST_METRIC2 = metric();
+    private static final String TEST_METRIC3 = metric();
 
     @BeforeClass
     public void prepareData() throws Exception {
         prepareApplyTestData(TEST_METRIC);
+
+        String TEST_ENTITY = entity();
```
GregoryKutuzov: add final modificator

PR #229, opened 2017-01-25
Bugfix: custom value deserializer and text samples mocks

unrealwork:
# Description
- [x] Custom value deserializer `ValueDeserializer`, to deserialize `"NaN"` value as `null` 

- [x] Text samples in  `Mocks` class

- [x] Added test for this deserialization:  `testSeriesQueryWithSampleWithoutValue`


unrealwork:
Is it correct deserialization?

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
```diff
@@ -615,6 +615,18 @@ public void testExactMatchIgnoresReservedVersioningTags() throws Exception {
         assertEquals("Number of received versions mismatched", insertedVersionsCount, receivedVersionsCount);
     }
 
+    @Test
+    public void testSeriesQueryWithSampleWithoutValue() throws Exception {
```
GregoryKutuzov: with text sample? =)

PR #230, opened 2017-01-26
DEV: models and methods to src

unrealwork:


pom.xml
```diff
@@ -132,6 +132,28 @@
 
     <build>
         <plugins>
+
+            <plugin>
+                <groupId>org.jacoco</groupId>
+                <artifactId>jacoco-maven-plugin</artifactId>
+                <version>${jacoco.version}</version>
```
rmakulov: Why do we need it?

src/test/java/com/axibase/tsd/api/transport/tcp/TCPSender.java
```diff
@@ -1,93 +0,0 @@
...
-            throw new Exception("Fail to check inserted command");
-    }
-
-    public void sendCheck(String command, long sleepDuration) throws Exception {
-        setCommand(command);
-        boolean successed = sendDebugMode(sleepDuration);
-        if (!successed)
-            throw new Exception("Fail to check inserted command");
-    }
-}
```
rmakulov: We got rid of this?

PR #234, opened 2017-01-31
SQL: keywords are case-sensitive  #3843

RawrLion:
added tests for #3843
TESTS FAIL, BECAUSE OF UNRESOLVED ISSUE


RawrLion:
Now everything is fine, work is done

PR #235, opened 2017-02-02
SQL: ORDER BY not applied when using column alias #3840

RawrLion:
added tests for #3840
**DON'T BE CONFUSED BY THE NAME OF THIS PULL REQUEST!**
Name was taken from ticket's name, but the problem of ORDER BY malfunction wasn't connected to column aliases.
The main problem was ordering by column with aggregation function that have parameters from multiple metrics -- like SUM(table1.value + table2.value).

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/clause/orderby/OrderByAggregatedValuesFromMultipleTables.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/clause/orderby/OrderByAggregatedValuesFromMultipleTables.java
@@ -13,8 +13,8 @@ import org.testng.annotations.Test;
 import java.util.ArrayList;
 import java.util.List;
 
-import static com.axibase.tsd.api.util.Util.TestNames.entity;
-import static com.axibase.tsd.api.util.Util.TestNames.metric;
+import static com.axibase.tsd.api.util.TestUtil.TestNames.entity;
+import static com.axibase.tsd.api.util.TestUtil.TestNames.metric;
 
 public class OrderByAggregatedValuesFromMultipleTables extends SqlTest {
     private static final String TEST_ENTITY = entity();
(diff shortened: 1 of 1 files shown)


src/test/java/com/axibase/tsd/api/method/sql/clause/orderby/OrderByAggregatedValuesFromMultipleTables.java
```diff
@@ -0,0 +1,104 @@
...
+    /**
+     * #3840
+     */
+    @Test
+    public void testOrderByColumnWithOneAggregationOfDifferentMetrics() {
+        String sqlQuery = String.format(
+                "SELECT SUM(t2.value-t1.value) AS col " +
+                        "FROM '%s' t1 JOIN '%s' t2 " +
+                        "GROUP BY t1.PERIOD(2 MILLISECOND) " +
+                        "ORDER BY col DESC",
```
GregoryKutuzov: shouldn't we add cases with inverse order?

src/test/java/com/axibase/tsd/api/method/sql/clause/orderby/OrderByAggregatedValuesFromMultipleTables.java
```diff
@@ -0,0 +1,156 @@
...
+    private static final String TEST_ENTITY = entity();
+    private static final String TEST_METRIC_1 = metric();
+    private static final String TEST_METRIC_2 = metric();
+
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        Registry.Entity.register(TEST_ENTITY);
+        Registry.Metric.register(TEST_METRIC_1);
+        Registry.Metric.register(TEST_METRIC_2);
+
```
rmakulov: Please, exclude manual registration by using another Series constructor further.

PR #244, opened 2017-02-13
Data API: series query returns no text value with last=true #3860

alexey-kravtsov:
added tests for #3860, SeriesQuery extended with "last" flag


GregoryKutuzov:
please, resolve conflicts

PR #256, opened 2017-03-15
CAST TEST, add format checks

VeselovAlex:


src/test/java/com/axibase/tsd/api/method/sql/function/string/CastTest.java
```diff
@@ -207,4 +216,124 @@ public void testCastHaving() {
...
+         assertEquals("string", resultTable.getColumnMetaData(0).getDataType());
+     }
+
+    /**
+     * #3770
+     */
+    @Test(dataProvider = "castNumberArgumentsProvider")
+    public void testCastNumberAsStringPassedToStringFunction(String castArgument) throws Exception {
+        Series series = castNumberAsStringSeries;
+        String sqlQuery = "SELECT CONCAT('foo', CAST("+castArgument+" AS string)) FROM '" + series.getMetric() + "'";
```
rmakulov: What if we use format here?

src/test/java/com/axibase/tsd/api/method/sql/function/string/CastTest.java
```diff
@@ -207,4 +216,124 @@ public void testCastHaving() {
...
+     * #3770
+     */
+     @Test(dataProvider = "castNumberArgumentsProvider")
+     public void testCastNumberAsStringApplied(String castArgument) throws Exception {
+         Series series = castNumberAsStringSeries;
+         String sqlQuery = "SELECT CAST("+castArgument+" AS string) FROM '" + series.getMetric() + "'";
+
+         StringTable resultTable = SqlMethod.queryTable(sqlQuery);
+
+         assertEquals("string", resultTable.getColumnMetaData(0).getDataType());
```
rmakulov: Please, add assert message to describe error if it happens.

PR #258, opened 2017-03-17
Test for concat in where

vtols:


src/test/java/com/axibase/tsd/api/method/sql/function/string/ConcatTest.java
```diff
@@ -127,4 +127,23 @@ public void testConcatWordAndTwoNumbers() throws Exception {
...
     }
+
+
+    /**
+     * #4017
+     */
+    @Test
+    public void testConcatInWhere() {
+        String sqlQuery = String.format(
+                "SELECT entity FROM '%s' WHERE entity = CONCAT('%s', '')",
```
rmakulov: Please, add a test with not empty string in the second argument and some other cases that might be used including not the only string.

vtols:
There's `testApply` function already, but it's name is not quite clear, maybe it should be renamed?

PR #261, opened 2017-03-17
Fixed slow test

alexey-kravtsov:
Fixed slow test for bug #3890


src/test/java/com/axibase/tsd/api/method/sql/response/SqlLargeDataTest.java
```diff
@@ -37,31 +43,44 @@ public static void initialize() throws Exception {
...
-    }
 
-    /**
-     * #3890
-     */
-    @Test
-    public void testQueryLargeData() {
-        String sqlQuery = String.format("SELECT COUNT(value) FROM '%s'", METRIC_NAME);
+        // wait for atsd insert series
+        Thread.sleep(30000);
```
rmakulov: Please, implement wait for insert in another way like in SeriesMethod.

unrealwork: @alexey-kravtsov 
I suggest to implement it by `Check`, [example](https://github.com/axibase/atsd-api-test/blob/master/src/test/java/com/axibase/tsd/api/transport/tcp/TCPSenderTest.java). You can check that metric contains the right amount of series. And you can send a collection of commands by `TCPSender`. But be careful, that ATSD can handle commands that length less than **128*1024** symbols.

alexey-kravtsov: @unrealwork 
All requests are splitted at chunks by 500 series, so command length won't be more than 128*1024 symbols.

PR #263, opened 2017-03-17
SQL:: Add tests for grouping by text value (#4002)

VeselovAlex:


src/test/java/com/axibase/tsd/api/method/sql/clause/groupby/GroupByText.java
```diff
@@ -0,0 +1,111 @@
...
+            } else {
+                sample = new Sample(String.format(ISO_PATTERN, minutes), new BigDecimal(1), text);
+            }
+            DEFAULT_SERIES.addData(sample);
+            minutes += 5;
+        }
+        SeriesMethod.insertSeriesCheck(DEFAULT_SERIES);
+    }
+
+    @Test
```
alexey-kravtsov: Please add issue number before all methods

src/test/java/com/axibase/tsd/api/method/sql/clause/groupby/GroupByText.java
```diff
@@ -0,0 +1,111 @@
...
+    public void testGroupByText() throws Exception {
+        String query = String.format(GROUP_TEXT_BY, "text");
```
alexey-kravtsov: I think code will be more clear if you put SQL query right into the method, not in class field. This applies to all methods

PR #266, opened 2017-03-20
SQL: GROUP BY entity.label doesn't work #3855

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/clause/groupby/GroupByCaseExpression.java
```diff
@@ -115,4 +124,24 @@ public void testGroupByColumnAlias() {
 
         assertSqlQueryRows("CASE in SELECT and GROUP BY gives wrong result", expectedRows, sqlQuery);
     }
+
+    /**
+     * #3912
```
GregoryKutuzov: wrong issue number

src/test/java/com/axibase/tsd/api/method/sql/clause/groupby/GroupByCaseExpression.java
```diff
@@ -115,4 +124,24 @@ public void testGroupByColumnAlias() {
...
+     */
+    @Test
+    public void testGroupByEntityLabel() {
```
GregoryKutuzov: bad test, add aggregation function and expect several lines as response.

PR #269, opened 2017-03-22
SQL: Add metric fields #3882

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectMetricFieldsTest.java
```diff
@@ -0,0 +1,53 @@
...
+                        "m.metric.filter,\n" +
+                        "m.metric.lastInsertTime,\n" +
+                        "m.metric.retentionIntervalDays,\n" +
+                        "m.metric.versioning,\n" +
+                        "m.metric.minValue,\n" +
+                        "m.metric.maxValue,\n" +
+                        "m.metric.invalidValueAction\n" +
+                 "FROM '%1s' m",
+                TEST_SERIES.getMetric()
+        );
```
rmakulov: What about metric.counter?

vtols: It is not available yet

src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectMetricFieldsTest.java
```diff
@@ -0,0 +1,53 @@
...
+                        "m.metric.versioning,\n" +
+                        "m.metric.minValue,\n" +
+                        "m.metric.maxValue,\n" +
+                        "m.metric.invalidValueAction,\n" +
+                        "m.metric.counter\n" +
+                 "FROM '%1s' m",
+                TEST_SERIES.getMetric()
+        );
+
+        String[][] expectedRows = {
```
vtols: Maybe it would be more readable with data provider or something similar, consisting of pairs like (field name, expected result)?

src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectMetricFieldsTest.java
```diff
@@ -0,0 +1,53 @@
...
+                { "null", "null", "LINEAR", "null", "FLOAT", "MILLISECONDS",
```
vtols: Maybe it would be better to create data provider or something similar, consisting of pairs (field name, expected result)?

PR #270, opened 2017-03-23
Test for CASE with value

vtols:


src/test/java/com/axibase/tsd/api/method/sql/operator/CaseValueTest.java
```diff
@@ -0,0 +1,178 @@
...
+        SeriesMethod.insertSeriesCheck(s1);
+
+        Series s2 = new Series(entity(), METRIC_NAME_B);
+        s2.addData(new TextSample("2017-01-01T09:30:00.000Z", "a"));
+        s2.addData(new TextSample("2017-01-02T09:30:00.000Z", "b"));
+        s2.addData(new TextSample("2017-01-03T09:30:00.000Z", "c"));
+        s2.addData(new TextSample("2017-01-04T09:30:00.000Z", "d"));
+        s2.addData(new TextSample("2017-01-05T09:30:00.000Z", "e"));
+        s2.addData(new TextSample("2017-01-06T09:30:00.000Z", "f"));
+        SeriesMethod.insertSeriesCheck(s2);
```
unrealwork: I believe that you can do only one call of `SeriesMethod.insertSeriesCheck` method.

src/test/java/com/axibase/tsd/api/method/sql/operator/CaseValueTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        assertSqlQueryRows("CASE <value> for strings gives wrong result", expectedRows, sqlQuery);
+    }
+
+    /**
+     * 4021
+     */
+    @Test
+    public void testCaseNoFrom() {
+        String sqlQuery = "SELECT CASE WHEN 0 < 1 THEN 1 ELSE 0 END\n";
```
unrealwork: What is the reason of the line break?

vtols: Strange, I don't know how it appeared here, it is not intended

vtols:
Not quite sure about package for this test

PR #271, opened 2017-03-23
Set fixed timezone

vtols:


src/test/java/com/axibase/tsd/api/method/sql/function/dateformat/DateFormatInsideClausesTest.java
```diff
@@ -22,14 +22,15 @@ public static void prepareData() throws Exception {
...
-                new Sample("2017-02-10T07:00:00.000Z", "12"),
+                new Sample("2017-02-10T08:00:00.000Z", "12"),
                 new Sample("2017-02-12T12:00:00.000Z", "13"),
                 new Sample("2017-02-11T12:00:00.000Z", "14"),
                 new Sample("2017-02-09T12:00:00.000Z", "15")
                 )
         );
 
         SeriesMethod.insertSeriesCheck(series);
+        System.out.println("OK");
```
unrealwork: @vtols I think it's not necessary ;)

vtols: Oops

PR #274, opened 2017-03-24
NPE when aggregating metrics with decimal datatype #3881

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/function/aggregation/AggregationChangedDatatypeTest.java
```diff
@@ -0,0 +1,46 @@
...
+        Series series = new Series();
+        series.setEntity(entity.getName());
+        series.setMetric(metric.getName());
+        series.addData(SAMPLE);
+        series.addTag("tag1", "1");
+
+        SeriesMethod.insertSeriesCheck(series);
+
+        metric.setDataType(DataType.DECIMAL);
+        MetricMethod.createOrReplaceMetric(metric);
```
GregoryKutuzov: do we need to call compaction here?

alexey-kravtsov: No, we don't need to call compaction. Error occurred right after data type change, not after change and compaction

PR #275, opened 2017-03-24
SQL:: Add tests for metric fields in JOIN  (#4027)

VeselovAlex:


src/test/java/com/axibase/tsd/api/method/sql/clause/join/JoinWithTags.java
```diff
@@ -255,4 +260,57 @@ public void testDifferentTagsAfterJoinUsingEntity() {
...
+    /**
+     * #4027
+     */
+    @Test
+    public void testJoinKeepsMetricTags() throws Exception {
+        Series series = Mocks.series();
+        Metric metric = new Metric();
+        metric.setName(series.getMetric());
+        Map<String, String> tags = new HashMap<>();
+        tags.put("foobar", "foo");
```
alexey-kravtsov: You can use Mocks.TAGS and check for ("value", "tag=value")

PR #277, opened 2017-03-27
SQL: column alias in CAST and ROW_NUMBER functions

vtols:
Can't choose right package for this test


src/test/java/com/axibase/tsd/api/method/sql/clause/with/ColumnAliasInFunctionTest.java
```diff
@@ -0,0 +1,68 @@
...
+        };
+
+        assertSqlQueryRows("Wrong result with alias inside ROW_NUMBER", expectedRows, sqlQuery);
+    }
+
+    /**
+     * 3842
+     */
+    @Test
+    public void testAliasInRowNumber() {
```
alexey-kravtsov: I think method name should be "testAliasInCastFunction" or similar

PR #284, opened 2017-03-30
SQL: string data type in BETWEEN operator

vtols:


src/test/java/com/axibase/tsd/api/method/sql/operator/SqlBetweenStringTest.java
```diff
@@ -0,0 +1,112 @@
...
+        String[][] expectedRows = {{"3", "a"}};
+
+        assertSqlQueryRows("Wrong result with equal string limits in BETWEEN operator", expectedRows, sqlQuery);
+    }
+
+    /**
+     * #3991
+     */
+    @Test
+    public void testBetweenAllString() {
```
alexey-kravtsov: It seems like this test and "testBetweenSimpleString" doing almost the same. I think this test can be removed

vtols: It is very similar, but I think it is not critical if leave it here

PR #289, opened 2017-04-03
SQL: NPE in WHERE clause IN list (#4047)

vtols:


src/test/java/com/axibase/tsd/api/method/sql/clause/where/WhereTagInListTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import static com.axibase.tsd.api.util.TestUtil.TestNames.entity;
+import static com.axibase.tsd.api.util.TestUtil.TestNames.metric;
+
+public class WhereTagInListTest extends SqlTest {
+    private static final String METRIC_NAME = metric();
+
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        String entityName = entity();
```
GregoryKutuzov: Move to constant as metric

PR #293, opened 2017-04-05
SQL: NPE in OUTER JOIN with partitioning

vtols:


src/test/java/com/axibase/tsd/api/method/sql/clause/join/OuterJoinTagsExternalTest.java
```diff
@@ -0,0 +1,60 @@
...
+    public void testOuterJoinTagsExternal() {
+        String sqlQuery = String.format(
+                "SELECT '%s'.tags FROM '%s' OUTER JOIN '%s' " +
+                        "OPTION (ROW_MEMORY_THRESHOLD 0)",
+                METRIC_NAME1,
+                METRIC_NAME1,
+                METRIC_NAME2
+        );
+
+        String[][] expectedRows = {{"null"}};
```
GregoryKutuzov: what is the use case of this query?

vtols: See this comment https://nur.axibase.com:41791/redmine/issues/4058#note-3

PR #294, opened 2017-04-05
SQL: LAG/LEAD function raise syntax error in BETWEEN operator

vtols:


src/test/java/com/axibase/tsd/api/method/sql/function/rows/FunctionsInsideBetweenOperatorTest.java
```diff
@@ -0,0 +1,60 @@
...
+    @Test
+    public void testSqrtInsideBetweenOperator() {
+        String sqlQuery = String.format(
+                "SELECT value FROM '%s' WHERE value BETWEEN sqrt(value) and value",
+                METRIC_NAME
+        );
+
+        String[][] expectedRows = {{"1"}, {"2"}, {"4"}, {"3"}, {"5"}, {"6"}};
+
+        assertSqlQueryRows("Incorrect result with lag/lead in BETWEEN operator",
```
GregoryKutuzov: change error message

PR #299, opened 2017-04-07
SQL: Incorrect result of max_value_time function

vtols:


vtols:
Note for me: maybe also add/update tests for `COUNT`

PR #301, opened 2017-04-10
Fixed test api for #3952 and #4088

alexey-kravtsov:


src/main/java/com/axibase/tsd/api/model/series/Sample.java
```diff
@@ -5,12 +5,20 @@
...
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
 import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import org.apache.commons.lang3.time.DateUtils;
 
 import java.math.BigDecimal;
+import java.text.ParseException;
 import java.util.Date;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Sample {
```
rodionos: Are optionals supported by whatever date parse is used, e.g. [.SSS]?

alexey-kravtsov: There are two date formats supported - with and without milliseconds (yyyy-MM-dd'T'HH:mm:ss.SSSXXX and yyyy-MM-dd'T'HH:mm:ssXXX). Is it required to add any other formats?

alexey-kravtsov: Yes, there are not all formats supported. Will be fixed in next commit

axibase-git:
My initial comment was to not add new formats but use format that supports "optionals", such as ":ss[.SSS]Z"?
It is better to use one format that can parse date in different ISO formats. Perhaps, there is a built-in format?

On Apr 10, 2017, at 6:15 PM, alexey-kravtsov <notifications@github.com> wrote:

@alexey-kravtsov commented on this pull request.

In src/main/java/com/axibase/tsd/api/model/series/Sample.java <https://github.com/axibase/atsd-api-test/pull/301#discussion_r110683366>:

>  
 import java.math.BigDecimal;
+import java.text.ParseException;
 import java.util.Date;
 
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public class Sample {
Yes, there are not all formats supported. Will be fixed in next commit

src/test/java/com/axibase/tsd/api/method/sql/response/SqlLargeDataTest.java
```diff
@@ -23,7 +24,7 @@
      * #3890
      */
     @Test
-    public void testQueryLargeData() throws IOException, InterruptedException {
+    public void testQueryLargeData() throws Exception {
```
rmakulov: Why have you changed exception?

alexey-kravtsov:
Changes already in master

PR #307, opened 2017-04-17
Entity and metric fields query test #4117

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectEntityFieldsTest.java
```diff
@@ -0,0 +1,79 @@
...
+public class SqlSelectEntityFieldsTest extends SqlTest {
+
+    private static final String TEST_ENTITY = entity();
+    private static final String TEST_METRIC = metric();
+
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        Registry.Metric.register(TEST_METRIC);
+
+        Entity entity = new Entity(TEST_ENTITY);
```
VeselovAlex: Why not to use Mocks.entity()?

alexey-kravtsov: Because I use entity name further in entityFieldsProvider

src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectMetricFieldsTest.java
```diff
@@ -1,79 +1,150 @@
...
 public class SqlSelectMetricFieldsTest extends SqlTest {
-
-    private static final Series TEST_SERIES = Mocks.series();
+    private static final String TEST_METRIC = metric();
 
     @BeforeClass
     public static void prepareData() throws Exception {
-        SeriesMethod.insertSeriesCheck(TEST_SERIES);
+        Metric metric = new Metric(TEST_METRIC, Mocks.TAGS);
+        metric.setName(TEST_METRIC);
```
alexey-kravtsov: Yes, because I set all metric fields and check them in test

PR #308, opened 2017-04-17
SQL: Incorrect first values interpolation #4069

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/function/interpolate/InterpolationBoundaryValuesTest.java
```diff
@@ -0,0 +1,114 @@
...
+        String sqlQuery = String.format(
+                "SELECT value " +
+                "FROM '%s' " +
+                "WHERE datetime BETWEEN '2011-12-31T00:00:00Z' AND '2012-01-03T00:00:00Z' " +
+                "      OR datetime BETWEEN '2012-01-06T00:00:00Z' AND '2012-01-10T00:00:00Z' " +
+                "WITH INTERPOLATE(1 DAY, PREVIOUS, INNER, NAN)",
+                TEST_METRIC);
+
+        String[][] expectedRows = {
+                {"NaN"}, {"1"}, {"2"}, {"3"}, {"NaN"}, {"7"}, {"8"}, {"9"}, {"9"}
```
vtols: @VeselovAlex, maybe we should beautify rows also here?

src/main/java/com/axibase/tsd/api/model/series/Sample.java
```diff
@@ -54,21 +53,26 @@ public Sample(Date d, String v) {
         this.v = new BigDecimal(String.valueOf(v));
     }
 
-    public Sample(Long t, BigDecimal v) {
```
vtols: Can you explain please, why did you remove this constructor?

PR #316, opened 2017-04-24
SQL: JOIN with atsd_series (#4089)

vtols:


src/test/java/com/axibase/tsd/api/method/sql/clause/join/JoinWithAtsdSeriesTest.java
```diff
@@ -0,0 +1,169 @@
...
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        String entityName = TestNames.entity();
+
+        Registry.Entity.register(entityName);
+
+        Registry.Metric.register(METRIC_NAME1);
+        Registry.Metric.register(METRIC_NAME2);
+
+        Series series1 = new Series();
```
rodionos: Create a constructor new Series(String entity, String metric, String... tags) to make the data preparation code easier to read:

```java
Series series2 = new Series(entityName, METRIC_NAME1, "t1", "tag");
```

Refactor other tests to make prepareData easier to read.

Also, review your own tests and add inserted data in tabular format as comments. 
This will make tests more readable.
The query used to display inserted data can be adjusted based on what is relevant for the test, for example, exclude the **text** column or expand tags. 

```ls
/**
SELECT entity, metric, value, text, tags FROM 'm-1'

| datetime                 | entity | metric | value | text | tags                  | 
|--------------------------|--------|--------|-------|------|-----------------------| 
| 1970-01-01T00:00:00.000Z | e-1    | m-1    | 3.5   | null | null                  | 
| 2015-08-06T17:44:32.430Z | e-1    | m-1    | 14.5  | null | null                  | 
| 2015-08-06T17:43:05.430Z | e-1    | m-1    | 13.5  | null | $version_status=Test1 | 
| 2015-10-13T17:50:00.000Z | e-1    | m-1    | 0.0   | null | null                  | 
*/
```

vtols: The inconvenience with `Series(String entity, String metric)` constructor is that it registers entity and metric names passed to, and so these names can't be reused in other `Series` without setting them separately.
Maybe `Series(String entity, String metric, String... tags)` constructor wouldn't be ok? Because usages with odd number of arguments are incorrect in this case
What if we add class named something like `SeriesContext` that will let us create different Series that share entity and metric? And make `addTag` to return `this`?
Then we can write
```java
...
SeriesContext ctx = new SeriesContext()
Series s1 = ctx.create(entityName, metricName).addTag("t1", "tag");
...
```

rodionos: The registration code is being refactored now by @rmakulov  and shall be completed momentarily. Once it is finished, the constructor will not perform registry checks. Instead the checks will be performed at insert time. As for varargs, the constructor will raise an exception of the tags are not submitted in pairs. varargs.length %2 must be 0. This will ensure that the tags are passed in pairs.

vtols: You propose to add comments with query+result, right?

src/test/java/com/axibase/tsd/api/method/sql/clause/join/JoinWithAtsdSeriesTest.java
```diff
@@ -0,0 +1,353 @@
...
+    private static final String METRIC_NAME3 = TestNames.metric();
+    private static final String METRIC_NAME4 = TestNames.metric();
+
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        String entityName = TestNames.entity();
+
+        Series series1 = new Series(entityName, METRIC_NAME1);
+        series1.addSamples(
+                new Sample("2017-01-01T12:00:00.000Z", "1"),
```
rmakulov: Why do we need here string for values?

src/test/java/com/axibase/tsd/api/method/sql/clause/join/JoinWithAtsdSeriesTest.java
```diff
@@ -0,0 +1,353 @@
...
+                        "FROM atsd_series t1 " +
+                        "JOIN USING ENTITY '%2$s' t2 " +
+                        "WHERE t1.metric = '%1$s'",
+                METRIC_NAME1,
+                METRIC_NAME2
+        );
+
+        String[][] expectedRows = {
+                {"3", "5"},
+                {"4", "6"}
```
rodionos: This order is not guaranteed. The database can return rows in any order if the ORDER BY clause is not included in the query. Fix other multi-row queries in this class.

PR #317, opened 2017-04-25
SQL: with interpolate execution slow in PREVIOUS mode

vtols:


vtols:
@rmakulov, what do you want to be described in comments?

PR #322, opened 2017-05-04
SQL: self-join error caused by IN metrics() function (#4149)

vtols:


src/test/java/com/axibase/tsd/api/method/sql/clause/where/WhereInNoSelfJoinTest.java
```diff
@@ -0,0 +1,176 @@
...
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        Registry.Entity.register(ENTITY_NAME);
+
+        Registry.Metric.register(METRIC_NAME1);
+        Series series1 = new Series();
+        series1.setEntity(ENTITY_NAME);
+        series1.setMetric(METRIC_NAME1);
+        series1.addData(new Sample("2016-10-04T01:58:12.000Z", "90.4"));
+        series1.addData(new Sample("2016-10-04T02:00:05.000Z", "97.7"));
```
rmakulov: What if we use setData and provide an explicit list of samples?

vtols: Maybe better add variable-length arguments list method `addData`?

PR #326, opened 2017-05-15
SQL: CAST to Number doesn't work with ISNULL with number #4182

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/function/string/CastTest.java
```diff
@@ -504,4 +504,276 @@ public void testImplicitCastInMathExpressionsRizesError() throws Exception {
...
+        String sqlQuery = String.format(
+                "SELECT CAST(ISNULL('5', '3') AS NUMBER) FROM '%s' ",
+                TEST_METRIC1_NAME
+        );
+
+        String[][] expectedRows = {
+                {"5"}
+        };
+
+        assertSqlQueryRows("CAST from string with ISNULL gives wrong result", expectedRows, sqlQuery);
```
rmakulov: Could you please replace (here and further) 'gives wrong result' with some description of expected behaviour that should take place?

PR #327, opened 2017-05-15
SQL: filter by tag not applied (Java 8) #4180

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/clause/where/SqlMerticTagsTest.java
```diff
@@ -0,0 +1,342 @@
...
+        Registry.Metric.register(TEST_METRIC);
+
+        String entity = entity();
+        Registry.Entity.register(entity);
+
+        String[] tagValues = new String[] { null, "value1", "value2", "value3", "otherValue" };
+
+        List<Series> seriesList = new ArrayList<>();
+        for (int i = 0; i < tagValues.length; i++) {
+            Series series = new Series();
```
rmakulov: Setters can be replaced with Series constructor now.

src/test/java/com/axibase/tsd/api/method/sql/clause/where/SqlMerticTagsTest.java
```diff
@@ -0,0 +1,337 @@
...
+import static com.axibase.tsd.api.util.TestUtil.TestNames.entity;
+
+public class SqlMerticTagsTest extends SqlTest {
+    private final String TEST_METRIC = metric();
+
+    @BeforeTest
+    public void prepareData() throws Exception {
+        String entity = entity();
+
+        String[] tagValues = new String[] { null, "value1", "value2", "value3", "otherValue" };
```
rmakulov: Why do we need null here?

src/test/java/com/axibase/tsd/api/method/sql/clause/where/SqlMerticTagsTest.java
```diff
@@ -0,0 +1,337 @@
...
+            Series series = new Series(entity, TEST_METRIC);
+            String tagValue = tagValues[i];
+            if (tagValue != null) {
+                series.addTag("tag", tagValue);
+            }
+            series.addData(new Sample(String.format("2017-01-01T00:0%S:00Z", i), i));
+
+            seriesList.add(series);
+        }
+
```
rmakulov: What about such loop instead?
```java
        for (int i = 1; i < tagValues.length; i++) {
            Series series = new Series(entity, TEST_METRIC,"tag", tagValues[i]);
            series.addData(new Sample(String.format("2017-01-01T00:0%S:00Z", i), i));
            seriesList.add(series);
        }
```

alexey-kravtsov: We need one series with no tag. This loop won't insert it

PR #328, opened 2017-05-16
SQL: period start time is incorrect for YEAR period and END_TIME alignment #4175

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/function/period/align/SqlPeriodAlignTest.java
```diff
@@ -140,4 +187,352 @@ public void testEndTimeExclusiveAlignment() {
...
+                                {"2008-01-01T00:00:00.005Z", "1"},
+                                {"2009-01-01T00:00:00.005Z", "2"},
+                                {"2010-01-01T00:00:00.005Z", "3"}
+                        }
+                },
+                //</editor-fold>
+        };
+    }
+
+
```
rodionos: Add tests for DAY+ periods using custom time zone during dates when daylight offset is changed, for example around last Sunday of March in Europe. PERIOD(1 DAY, END_TIME, "Europe/Vienna"). Note that PERIOD in this case is calculated as local time +/- N day, so 4-April 15:00 in local timezone - 14 days = 21 March 15:00 local time.

src/test/java/com/axibase/tsd/api/method/sql/function/period/align/SqlPeriodAlignTest.java
```diff
@@ -140,4 +194,424 @@ public void testEndTimeExclusiveAlignment() {
         );
         assertEquals(expectedTableRows, resultTableRows);
     }
+
+
+    @DataProvider(name = "periodsQueryParametersProvider")
+    public Object[][] providePeriodsQueryParameters() {
```
rodionos: This test is difficult to read. I'd like to be able to understand the input data, the query text, and the expected values. In this test, all 3 pieces are defined in different methods, and there is no explanation or hint on why the values should be returned as expected. Can we remove shared methods/provider to make tests as simple as possible: data, query, expected values.

src/test/java/com/axibase/tsd/api/method/sql/function/period/align/SqlPeriodAlignTest.java
```diff
@@ -140,4 +147,851 @@ public void testEndTimeExclusiveAlignment() {
...
+        };
+
+        assertSqlQueryRows(expectedRows, sqlQuery);
+    }
+
+    /**
+     * #4175
+     */
+    @Test
+    public void testGroupByPeriodSecondStartTime() throws Exception {
```
rmakulov: What if we add some sample before startTime?

src/test/java/com/axibase/tsd/api/method/sql/function/period/align/SqlPeriodAlignTest.java
```diff
@@ -140,4 +147,851 @@ public void testEndTimeExclusiveAlignment() {
...
+        };
+
+        assertSqlQueryRows(expectedRows, sqlQuery);
+    }
+
+    /**
+     * #4175
+     */
+    @Test
+    public void testGroupByPeriodSecondEndTime() throws Exception {
```
rmakulov: and add some sample after endTime?

rmakulov:
Please, try to make smaller pr in the future: one pr for one ticket.

src/test/java/com/axibase/tsd/api/method/sql/function/period/align/SqlPeriodAlignTest.java
```diff
@@ -14,25 +15,29 @@
...
-                    new Sample("2016-06-03T09:26:00.000Z", "8.1"),
-                    new Sample("2016-06-03T09:36:00.000Z", "6.0"),
-                    new Sample("2016-06-03T09:41:00.321Z", "19.0"),
-                    new Sample("2016-06-03T09:45:00.126Z", "19.0"),
-                    new Sample("2016-06-03T09:45:00.400Z", "17.0")
-            ));
-        }};
+    public void prepareDataSet() throws Exception {
+        insertSamples(
+                new Sample("2016-06-03T09:20:00.124Z", "16.0"),
```
rmakulov: Why do we use string for values?

rmakulov: I can't find where have you registered entity and metric.

rmakulov:
Please, resolve conflicts.

PR #329, opened 2017-05-22
SQL: Exception when multiple conditions are in parentheses #4195

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlSyntaxParenthesesTest.java
```diff
@@ -0,0 +1,131 @@
...
+import static com.axibase.tsd.api.util.TestUtil.TestNames.entity;
+import static com.axibase.tsd.api.util.TestUtil.TestNames.metric;
+
+public class SqlSyntaxParenthesesTest extends SqlTest {
+    private final String TEST_METRIC = metric();
+
+    @BeforeTest
+    public void prepareData() throws Exception {
+        String entity = entity();
+        Registry.Entity.register(entity);
```
rmakulov: Why do we here make separate entity and metric registration?

src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlSyntaxParenthesesTest.java
```diff
@@ -0,0 +1,126 @@
...
+public class SqlSyntaxParenthesesTest extends SqlTest {
+    private final String TEST_METRIC = metric();
+
+    @BeforeTest
+    public void prepareData() throws Exception {
+        String entity = entity();
+
+        Series series = new Series(entity, TEST_METRIC);
+        series.setData(Arrays.asList(
+                new Sample("2017-01-01T00:00:00Z", "0", "zero"),
```
rmakulov: Why do we use string for values?

PR #332, opened 2017-05-24
Series constructor with tags

vtols:


src/main/java/com/axibase/tsd/api/model/series/Series.java
```diff
@@ -35,6 +35,19 @@ public Series(String entity, String metric) {
...
     }
 
+    public Series(String entity, String metric, String... tags) {
+        this(entity, metric);
+
+        /* Tag name-value pairs */
+        if (tags.length % 2 != 0) {
+            throw new IllegalArgumentException("Tag name without value in arguments");
+        }
+
```
rodionos: Raise exception if either tag key or tag value is empty or null

src/main/java/com/axibase/tsd/api/model/series/Series.java
```diff
@@ -35,6 +35,26 @@ public Series(String entity, String metric) {
...
+        for (int i = 0; i > tags.length; i += 2) {
```
rmakulov: >?

src/main/java/com/axibase/tsd/api/model/series/Series.java
```diff
@@ -99,6 +120,12 @@ public void addData(Sample sample) {
         data.add(sample);
     }
 
+    public void addData(Sample... samples) {
```
rmakulov: It seems, that this method allows removing `public void addData(Sample sample)`.

PR #333, opened 2017-05-24
SQL: Incorrect result of OUTER JOIN USING ENTITY #4157

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/clause/join/SqlOuterJoinWithTagsTest.java
```diff
@@ -71,4 +69,60 @@ public void testJoinUsingEntityWithTags() {
 
         assertSqlQueryRows("JOIN USING ENTITY with tags gives wrong result", expectedRows, sqlQuery);
     }
+
+    /**
+     * #4157
+     */
+    @Test
+    public void testOuterJoinUsingEntity() throws Exception {
+        Series series1 = new Series();
```
vtols: Try to use the new constructor please, when this PR will be merged https://github.com/axibase/atsd-api-test/pull/332/files.

src/test/java/com/axibase/tsd/api/method/sql/clause/join/SqlOuterJoinWithTagsTest.java
```diff
@@ -71,4 +69,60 @@ public void testJoinUsingEntityWithTags() {
...
+        series1.setEntity(TEST_ENTITY_NAME);
+        series1.setMetric(TEST_METRIC1_NAME);
+        series1.addData(new Sample("2017-01-03T12:00:00.000Z", "3"));
+        series1.addTag("t1", "tag");
+
+        Series series2 = new Series();
+        series2.setEntity(TEST_ENTITY_NAME);
+        series2.setMetric(TEST_METRIC1_NAME);
+        series2.setData(Arrays.asList(
```
vtols: Also, you will be able to use `addData(Sample...)`

src/test/java/com/axibase/tsd/api/method/sql/clause/join/SqlOuterJoinWithTagsTest.java
```diff
@@ -71,4 +69,60 @@ public void testJoinUsingEntityWithTags() {
...
+                new Sample("2017-01-02T12:00:00.000Z", "2"),
+                new Sample("2017-01-04T12:00:00.000Z", "4")
+        ));
+
+        Series series3 = new Series();
+        series3.setEntity(TEST_ENTITY_NAME);
+        series3.setMetric(TEST_METRIC2_NAME);
+        series3.addData(new Sample("2017-01-03T12:00:00.000Z", "5"));
```
rmakulov: We should be able to use Series constructor with tags now.

src/test/java/com/axibase/tsd/api/method/sql/clause/join/SqlOuterJoinWithTagsTest.java
alexey-kravtsov: I can't use this constructor because entity and metric are registering in prepareData method for other test, I just using them. If I use this constructor, test won't pass because of registry error

src/test/java/com/axibase/tsd/api/method/sql/clause/join/SqlOuterJoinWithTagsTest.java
alexey-kravtsov: Series constructor with tags registering entity and metrics, so I can't use it in this test (my comment above)

src/test/java/com/axibase/tsd/api/method/sql/clause/join/SqlOuterJoinWithTagsTest.java
rmakulov: Could we move series insert to prepareBlock like in the other tests?

src/test/java/com/axibase/tsd/api/method/sql/clause/join/SqlOuterJoinWithTagsTest.java
```diff
@@ -71,4 +69,60 @@ public void testJoinUsingEntityWithTags() {
...
+        series1.setEntity(TEST_ENTITY_NAME);
+        series1.setMetric(TEST_METRIC1_NAME);
+        series1.addSamples(new Sample("2017-01-03T12:00:00.000Z", "3"));
+        series1.addTag("t1", "tag");
+
+        Series series2 = new Series();
+        series2.setEntity(TEST_ENTITY_NAME);
+        series2.setMetric(TEST_METRIC1_NAME);
+        series2.addSamples(
+                new Sample("2017-01-02T12:00:00.000Z", "2"),
```
rmakulov: Why do we use string for values?

PR #341, opened 2017-05-31
New replacement table registration

RawrLion:
replacement table tuned for new Registry functionality


src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -53,4 +60,29 @@ private static String squashMapIntoString(Map<String, String> map) {
...
+        InputStream bodyStream = (InputStream) response.getEntity();
+        String body;
+        try {
+            body = IOUtils.toString(bodyStream, "UTF-8");
+        }
+        catch (IOException ignored) {
+            throw new IllegalStateException("Error in decoding stream to string");
+        }
+
+        return body.contains(replacementTable);
```
rmakulov: Do you think that this kind of checking is enough here?

PR #353, opened 2017-06-14
SQL: error when using non-overlapping datetime filter by different metrics #4258

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/clause/join/JoinWithDatetime.java
```diff
@@ -394,4 +397,42 @@ public void testFullTimeSyntaxInOrderBy() {
...
+                        "WHERE m1.time >= 1483228800000 AND m1.time <= 1483315200000 " +
+                        "   AND m2.time >= 1483660800000 AND m2.time <= 1483747200000",
+                TEST_METRIC_1, TEST_METRIC_2
+        );
+
+        Response response = SqlMethod.queryResponse(sqlQuery);
+
+        assertBadRequest("Invalid date conditions", response);
+    }
 }
```
rodionos: The error is raised on non-intersecting intervals. Add test that intersecting intervals work.

PR #354, opened 2017-06-14
SQL: Missing metric expression error when using complex metric filter #4259

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/examples/select/SqlSelectFromAtsdSeriesTest.java
```diff
@@ -102,4 +102,20 @@ public void testExample3() {
...
+
+    /**
+     * #4259
+     */
+    @Test
+    public void testErrorOnComplexMetricFilter() {
+        String sqlQuery = String.format(
+                "SELECT entity, metric, datetime, value " +
+                        "FROM atsd_series " +
+                        "WHERE (metric = '%s') OR (metric = '%s' AND entity = '%s')",
```
rodionos: What additional metric conditions should raise this error "Invalid metric expression". Test for more conditions, add a comment describing when the error will be raised.

src/test/java/com/axibase/tsd/api/method/sql/examples/select/SqlSelectFromAtsdSeriesTest.java
```diff
@@ -102,4 +102,99 @@ public void testExample3() {
...
     }
+
+    /**
+     * #4259
+     */
+    @Test
+    public void testErrorOnComplexMetricAndEntityFilter() {
+        String sqlQuery = String.format(
+                "SELECT entity, metric, datetime, value " +
+                        "FROM atsd_series " +
```
rmakulov: It seems that there is no need to list the columns explicitly and * will be enough.

PR #357, opened 2017-06-16
SQL: Change logic rules according to three-valued logic #4286

RawrLion:
added tests for #4286

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
@@ -30,12 +30,10 @@ public class ThreeValuedLogicTest extends SqlTest {
         seriesList.add(series);
 
         series = new Series(ENTITY_NAME, METRIC_NAME_2);
-//        series.setEntity(ENTITY_NAME);
         series.addSamples(new Sample(ISO_TIME, 2));
         seriesList.add(series);
 
         series = new Series(ENTITY_NAME, METRIC_NAME_3);
-//        series.setEntity(ENTITY_NAME);
         series.addSamples(new TextSample(ISO_TIME, "hello"));
         seriesList.add(series);
 
--- a/src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
@@ -29,13 +29,13 @@ public class ThreeValuedLogicTest extends SqlTest {
         series.addSamples(new Sample(ISO_TIME, 1));
         seriesList.add(series);
 
-        series = new Series(null, METRIC_NAME_2);
-        series.setEntity(ENTITY_NAME);
+        series = new Series(ENTITY_NAME, METRIC_NAME_2);
+//        series.setEntity(ENTITY_NAME);
         series.addSamples(new Sample(ISO_TIME, 2));
         seriesList.add(series);
 
-        series = new Series(null, METRIC_NAME_3);
-        series.setEntity(ENTITY_NAME);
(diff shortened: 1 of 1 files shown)


src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @BeforeTest
+    private static void prepareData() throws Exception {
+        Series series = new Series(entity(), METRIC_NAME);
+        series.addSamples(SAMPLE);
+
+        SeriesMethod.insertSeriesCheck(series);
+    }
+
+    @DataProvider(name = "logicTableTestProvider")
```
rodionos: Add NULL checks to make sure that it's evaluation is the same as NaN.

src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
```diff
@@ -0,0 +1,84 @@
...
+                param
+        );
+        String actualValue = queryTable(query).getValueAt(0, 0);
+        String assertMessage = String.format("Incorrect result of %s function in SELECT clause", param);
+
+        assertEquals(assertMessage, expectedValue, actualValue);
+    }
+
+    /**
+     * #4286
```
rodionos: In addition to comparing explicit expression such as NaN = NaN, check the logic on column values from a JOIN:

```ls
series d:2017-05-16T00:00:00Z e:e-1 m:m-1=1 
series d:2017-05-16T00:00:00Z e:e-1 m:m-2=2
series d:2017-05-16T00:00:00Z e:e-1 x:m-3=hello
```

```sql
SELECT t1.value = t3.value, t1.tags IS NULL, t2.tags IS NULL, t1.tags = t2.tags 
  FROM 'm-1' t1 JOIN 'm-2' t2 JOIN 'm-3' t3
```

Note that t1.tags = t2.tags returns TRUE which is incorrect according to our Truth tables.

src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @BeforeTest
+    private static void prepareData() throws Exception {
+        List<Series> seriesList = new ArrayList<>();
+
+        Series series = new Series(ENTITY_NAME, METRIC_NAME_1);
+        series.addSamples(new Sample(ISO_TIME, 1));
+        seriesList.add(series);
+
+        series = new Series(null, METRIC_NAME_2);
```
rmakulov: Why do we need entity null here?

src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
```diff
@@ -0,0 +1,152 @@
...
+    public Object[][] provideJoinTestsData() {
+        return new Object[][]{
+                {"t1.tags IS NULL", "true"},
+                {"t2.tags IS NULL", "true"},
+                {"t3.tags IS NULL", "true"},
+                {"t1.value = t3.value", "null"},
+                {"t1.text = t3.text", "null"},
+                {"t1.tags = t2.tags", "null"},
+                {"t2.tags = t3.tags", "null"}
+        };
```
rmakulov: Could we compare other columns?

src/test/java/com/axibase/tsd/api/method/sql/clause/where/ThreeValuedLogicTest.java
RawrLion: Because I want to JOIN several metrics, and they should have same entity. But I can't use (entity, metric) constructor, because of new Registry

PR #358, opened 2017-06-19
SQL: NULL value returned for lastInsertTime (#4035)

vtols:


src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectMetricFieldsTest.java
```diff
@@ -60,4 +61,15 @@ public void testQueryMetricFields(String field) {
...
     }
+
+    @Test
+    public void testMetricLastInsertTimeNotNull() {
+        String sqlQuery = String.format("SELECT metric.lastInsertTime FROM '%s'",
+                TEST_SERIES.getMetric());
+
+        StringTable resultTable = queryTable(sqlQuery);
+
+        boolean ok = !"null".equals(resultTable.getValueAt(0, 0));
```
rmakulov: What if we check for the specific time?

src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectMetricFieldsTest.java
```diff
@@ -60,4 +61,15 @@ public void testQueryMetricFields(String field) {
         // check for row existence
         assertEquals(String.format("Error in metric field query (%s)", field), resultTable.getRows().size(), 1);
     }
+
```
rmakulov: Please, add issue number.

src/test/java/com/axibase/tsd/api/method/sql/clause/select/SqlSelectMetricFieldsTest.java
vtols: It would be nice, but how to do this?

PR #371, opened 2017-07-28
Tidy up Sample class

vtols:
* Introduce Lombok
* Replace messy constructors with static factory methods


src/test/java/com/axibase/tsd/api/method/sql/function/string/ConcatTest.java
```diff
@@ -36,21 +34,21 @@ private static void prepareFunctionalConcatTestData() throws Exception {
 
         Series series = new Series(TEST_ENTITY, TEST_METRIC1);
         series.addSamples(
-                new TextSample("2016-06-03T09:19:18.000Z", ""), // NaN value
-                new Sample("2016-06-03T09:20:18.000Z", 3),
-                new Sample("2016-06-03T09:21:18.000Z", new BigDecimal("3.10")),
-                new Sample("2016-06-03T09:22:18.000Z", new BigDecimal("3.14")),
-                new Sample("2016-06-03T09:23:18.000Z", new BigDecimal("3.1415"))
+                Sample.ofDateText("2016-06-03T09:19:18.000Z", ""), // NaN value
```
alexey-kravtsov: Can you use ofDate there? It also passes value = null

PR #372, opened 2017-08-02
Tests for #3845

vtols:


src/test/java/com/axibase/tsd/api/method/sql/clause/select/SelectRowNumberTest.java
```diff
@@ -0,0 +1,99 @@
...
+    public void testSelectRowNumberOrderBy() {
+        String sqlQuery = String.format(
+                "SELECT value, row_number() FROM \"%s\" " +
+                        "WITH ROW_NUMBER(entity order by value) > 0 " +
+                        "ORDER BY row_number(), value",
+                METRIC_NAME
+        );
+
+        String[][] expectedResult = {
+                {"1", "1"},
```
raipc: Without reading the class from the beginning it's hard to understand why the expected values should be like these. Could you also add entity to SELECT statement?

src/test/java/com/axibase/tsd/api/method/sql/clause/select/SelectRowNumberTest.java
```diff
@@ -0,0 +1,123 @@
...
+    private static final String ENTITY_NAME1 = entity();
+    private static final String ENTITY_NAME2 = entity();
+    private static final String ENTITY_NAME3 = entity();
+    private static final String METRIC_NAME = metric();
+
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        Series series1 = new Series(ENTITY_NAME1, METRIC_NAME);
+        series1.addSamples(
+                new Sample("2017-06-20T00:00:00Z", 1),
```
rmakulov: It seems that constructor have to be replaced with appropriate static method because we merged #371.

PR #377, opened 2017-08-22
Tests for #4147

vtols:


src/test/java/com/axibase/tsd/api/method/sql/function/dateformat/TimeZoneFieldAsArgumentTest.java
```diff
@@ -0,0 +1,218 @@
...
+import java.time.temporal.ChronoUnit;
+import java.util.TimeZone;
+
+public class TimeZoneFieldAsArgumentTest extends SqlTest {
+    private static final String ENTITY_NAME = Mocks.entity();
+    private static final String METRIC_NAME1 = Mocks.metric();
+    private static final String METRIC_NAME2 = Mocks.metric();
+    private static final String METRIC_NAME3 = Mocks.metric();
+
+    private static final TimeZone ENTITY_ZONE = TimeZone.getTimeZone("Asia/Kathmandu");
```
raipc: The tests don't check that entity.timeZone is used, because Asia/Kathmandu is ATSD local timezone. Please, choose some other timezone for entity.

PR #381, opened 2017-08-24
Fixed concat and cast tests due to ATSD decimal format rounding mode changes

raipc:


raipc:
@rmakulov, please merge
https://travis-ci.org/axibase/atsd-api-test/builds/267895590

PR #388, opened 2017-09-04
pom.xml fixes

raipc:
Removed `jersey 1` dependency (jersey 2 is used)
Added `allure-java-commons` dependency for Issue annotation.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -17,7 +17,6 @@
         <javax.json.version>1.0.4</javax.json.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-        <jackson.version>2.7.3</jackson.version>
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -141,6 +134,14 @@
             <version>1.16.18</version>
             <scope>provided</scope>
         </dependency>
+
+        <!--for Issue annotation-->
```
vtols: How we will use it?

raipc: import io.qameta.allure.Issue;
@Issue("4000")

See the example: [atsd-jdbc-test-example](https://github.com/axibase/atsd-jdbc-test/blob/master/src/test/java/com/axibase/tsd/driver/jdbc/data/TagsManipulationTest.java#L25)
This approach takes less lines than comments and allows to include links to [generated report](https://github.com/axibase/atsd-jdbc-test#generate-report-and-run-on-localhost) (now reporting is not set up because of difficulties caused by parallel execution)

PR #401, opened 2017-09-20
Data API: series query - add support for tag expression #3915

alexey-kravtsov:


rodionos:
rebase and make sure status is OK

PR #405, opened 2017-09-25
Tests for #4581

vtols:


src/test/java/com/axibase/tsd/api/method/metric/MetricSeriesDateLimitTest.java
```diff
@@ -0,0 +1,137 @@
...
+                new MetricSeriesParameters()
+                        .setMinInsertDate(minDate)
+                        .setMaxInsertDate(maxDate);
+        MetricSeriesResponse[] responseMetrics;
+        try {
+            responseMetrics = MetricMethod.queryMetricSeries(METRIC_NAME, params)
+                    .readEntity(MetricSeriesResponse[].class);
+        } catch (Exception e) {
+            e.printStackTrace();
+            fail();
```
alexey-kravtsov: Please add fail cause here. You can use fail(String, Throwable) for this

PR #407, opened 2017-09-29
Tests for #4171

vtols:
#4602 should be resolved


vtols:
#4602 should be resolved

PR #410, opened 2017-10-04
Constant SAMPLE_TIME replaced with auto-generated

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/series/SeriesInsertTest.java
```diff
@@ -513,7 +513,7 @@ public void testMillisecondsUnsupported() throws Exception {
         String metricName = "m-iso-13";
 
         Series series = new Series(entityName, metricName);
-        series.addSamples(Sample.ofRawDateInteger(Mocks.MILLS_TIME.toString(), 0));
+        series.addSamples(Sample.ofRawDateInteger("1470009600000", 0));
```
vtols: It is a bit strange that we use timestamp here. How does it work?
https://github.com/axibase/atsd/blob/master/api/data/series/insert.md#value-object

alexey-kravtsov: This test checks for incorrect date string. Timestamp is passed as a date and test waits for error message

vtols: OK

src/test/java/com/axibase/tsd/api/util/Mocks.java
```diff
@@ -4,14 +4,18 @@
...
 import java.util.Date;
 import java.util.Map;
 
 public class Mocks {
     private static final TestNameGenerator NAME_GENERATOR = new TestNameGenerator();
 
-    public static final String ISO_TIME = "2017-08-01T00:00:00.000Z";
+    // generating date with current year because ATSD does not store messages older than year
+    public static final String ISO_TIME =
+            String.format("%s-08-01T00:00:00.000Z",
```
vtols: Maybe it is better to substitute the date with current year right in place where it is necessary?

vtols:
You can leave ISO_TIME unchanged

PR #413, opened 2017-10-05
Test for #4597

vtols:


src/test/java/com/axibase/tsd/api/method/sql/clause/where/WhereDatesWithJoinTest.java
```diff
@@ -0,0 +1,138 @@
...
+    @DataProvider
+    public Object[][] provideJoinTypeAndCondition() {
+        return new Object[][] {
+                {"JOIN", "t1.time = t2.time"},
+                {"JOIN", "t1.time = t2.datetime"},
+                {"JOIN", "t1.datetime = t2.datetime"},
+
+                {"FULL JOIN", "t1.time = t2.time"},
+                {"FULL JOIN", "t1.time = t2.datetime"},
+                {"FULL JOIN", "t1.datetime = t2.datetime"},
```
alexey-kravtsov: I think it would be useful to add OUTER JOIN / FULL OUTER JOIN join types

alexey-kravtsov: ok)

src/test/java/com/axibase/tsd/api/method/sql/clause/where/WhereDatesWithJoinTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        /* Same entity, no tags */
+        Series series3 = new Series(ENTITY_NAME1, LEFT_METRIC_NAME);
+        series3.addSamples(Sample.ofDateInteger("2017-09-02T00:00:00.000Z", 3));
+
+        Series series4 = new Series(ENTITY_NAME1, RIGHT_METRIC_NAME);
+        series4.addSamples(Sample.ofDateInteger("2017-09-02T00:00:00.000Z", 4));
+
+        /* Different entity, different tags */
+        Series series5 = new Series(ENTITY_NAME1, LEFT_METRIC_NAME);
```
rmakulov: Could we use here series constructor with tags directly?

PR #415, opened 2017-10-11
Search: API url #4404

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/series/SeriesSearchTest.java
```diff
@@ -1,47 +1,295 @@
...
+                new Filter<>("sst_11_entity_tag:sst_11_entity_value", resultRecord3),
+                new Filter<>("sst_other_metric_tag:sst_other_metric_value", resultRecord5)
         );
 
-        insertSeriesCheck(series);
         updateSearchIndex();
     }
 
+    private static SeriesSearchResultRecord createSeries(String prefix) throws Exception {
+        Entity entity = new Entity()
```
rmakulov: Please, use existing constructor here and further https://github.com/axibase/atsd-api-test/blob/kravtsov-4404/src/main/java/com/axibase/tsd/api/model/entity/Entity.java#L42

src/test/java/com/axibase/tsd/api/method/series/SeriesSearchTest.java
```diff
@@ -1,47 +1,295 @@
...
+                .setTags(Collections.singletonMap(prefix + "entity_tag", prefix + "entity_value"))
+                .setLabel(prefix + Mocks.LABEL)
+                .setInterpolationMode(InterpolationMode.PREVIOUS)
+                .setTimeZoneID("Europe/Moscow")
+                .setEnabled(true);
+
+        Metric metric = new Metric()
+                .setName(prefix + Mocks.metric())
+                .setTags(Collections.singletonMap(prefix + "metric_tag", prefix + "metric_value"))
+                .setLabel(prefix + Mocks.LABEL)
```
rmakulov: Please, use existing constructor here and further https://github.com/axibase/atsd-api-test/blob/kravtsov-4404/src/main/java/com/axibase/tsd/api/model/metric/Metric.java#L60

src/test/java/com/axibase/tsd/api/method/series/SeriesSearchTest.java
```diff
@@ -1,47 +1,295 @@
...
+        assertNotNull(entity.getLastInsertDate());
+
+        Metric metric = resultRecord.getMetric();
+        assertNotNull(metric);
+        assertNotNull(metric.getCreatedDate());
+        assertNotNull(metric.getLastInsertDate());
+
+        entity.setCreatedDate(null);
+        entity.setLastInsertDate(null);
+        assertEquals(entity, resultRecord4.getEntity());
```
rmakulov: Why do we need to explicitly set read-only fields?

src/test/java/com/axibase/tsd/api/util/Filters.java
```diff
@@ -0,0 +1,79 @@
...
+            for (Filter<T> secondFilter : filters2) {
+                String filter = String.format("(%s) OR (%s)", firstFilter.getExpression(), secondFilter.getExpression());
+                Set<T> expectedValues = Sets.union(firstFilter.getExpectedResultSet(), secondFilter.getExpectedResultSet());
+                result.add(new Filter<>(filter, expectedValues));
+            }
+        }
+
+        return result;
+    }
+
```
rmakulov: It seems that above 2 pairs of methods have the same body excluding first line, please, refactor.

src/test/java/com/axibase/tsd/api/method/series/SeriesSearchTest.java
alexey-kravtsov: Because I check entity and metrics for equality. If createdDate and lastInsertDate is different test will fail. I check this fields for existence (not null), then set them to null and compare with expected result

PR #418, opened 2017-10-11
SQL: allow aliases that are equal column names #4353

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/clause/select/SelectAliasSameAsColumnTest.java
```diff
@@ -0,0 +1,109 @@
...
+    @BeforeClass
+    public void prepareData() throws Exception {
+        Series series = Mocks.series();
+        TEST_METRIC = series.getMetric();
+
+        SeriesMethod.insertSeriesCheck(series);
+    }
+
+    @Issue("4353")
+    @Test
```
vtols: Use test descriptions. What each test tries ti check, what is different in both of tests

PR #436, opened 2017-10-24
Add tests for #4662

vtols:


src/test/java/com/axibase/tsd/api/util/TestNameGenerator.java
```diff
@@ -1,49 +1,52 @@
...
     private static final String API_METHODS_PACKAGE_NAME = "com.axibase.tsd.api";
-    private Map<String, Integer> dictionary;
-    private Class<org.testng.annotations.Test> TEST_ANNOTATION = org.testng.annotations.Test.class;
-
-    TestNameGenerator() {
-        this.dictionary = new HashMap<>();
-    }
+    private static final Class<Test> TEST_ANNOTATION = org.testng.annotations.Test.class;
 
+    private Map<String, Integer> prefixDictionary = new HashMap<>();
```
alexey-kravtsov: Please change to ConcurrentHashMap

src/test/java/com/axibase/tsd/api/method/sql/meta/MetaDataTest.java
```diff
@@ -304,8 +304,8 @@ public void testMetaNonEmptyTagsExpansion() throws Exception {
                 "entity",
                 "tags",
                 /* Do we always have this order? */
-                metricB + ".tags.t2",
-                metricA + ".tags.t1"
+                metricA + ".tags.t1",
+                metricB + ".tags.t2"
         };
```
rmakulov: Why do we change change the order?

src/main/java/com/axibase/tsd/api/model/series/Series.java
```diff
@@ -73,11 +72,37 @@ public Series copy() {
...
+        for (Sample sample : data) {
+            dataCopy.add(sample.copy());
+        }
+        copy.setSamples(dataCopy);
         copy.setTags(new HashMap<>(tags));
         return copy;
     }
 
+
+    public Series normalize() {
```
rmakulov: Could you please add a few words about the normalization meaning?

src/test/java/com/axibase/tsd/api/method/sql/meta/MetaDataTest.java
vtols: The order is unspecified, I don't know how to solve this problem. I proposed to fix the tags order, Alexander agreed, but Seregey declined this proposition.

src/main/java/com/axibase/tsd/api/model/series/Series.java
vtols: It is transforms series to for, in which we would get it back from ATSD, so lowercase metric, entity and tag names, t → d field replacement.

vtols: See also https://github.com/axibase/atsd-api-test/pull/441/files, Alexey checked out this change to his own PR.

PR #439, opened 2017-10-25
Search API: LastInsertDate and CreatedDate millis precision #4666

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/series/SeriesSearchTest.java
```diff
@@ -261,6 +265,31 @@ public void testFieldsTags() {
...
+
+        Response response = SeriesMethod.searchSeriesReturnResponse(query);
+        ObjectMapper mapper = new ObjectMapper();
+        JsonNode resultNode = mapper.readTree(response.readEntity(String.class));
+        SimpleDateFormat dateFormat = new SimpleDateFormat("YYYY-MM-dd'T'HH:mm:ss.SSS'Z'");
+        List<JsonNode> dateNodes = new ArrayList<>();
+        dateNodes.addAll(resultNode.findValues("createdDate"));
+        dateNodes.addAll(resultNode.findValues("lastInsertDate"));
+
+        assertEquals(dateNodes.size(),4, "Incorrect fields count");
```
vtols: Space `,4` :smiley:

src/main/java/com/axibase/tsd/api/method/series/SeriesMethod.java
```diff
@@ -120,6 +120,17 @@ public static SeriesSearchResult searchSeries(SeriesSearchQuery query) {
         return response.readEntity(SeriesSearchResult.class);
     }
 
+    public static Response searchSeriesReturnResponse(SeriesSearchQuery query) {
```
vtols: The name `searchSeriesReturnResponse` isn't good, especially because of `*Return*` part, think which name would describe better what the method does.

alexey-kravtsov: Name is similar to executeQueryReturnSeries, *Return* part describes difference between this method and searchSeries which returns SeriesSearchResult

vtols: I renamed `executeQueryReturnSeries` to `querySeriesAsList`. I think it is _really_ a bad name

PR #443, opened 2017-10-30
Data API: series query error #4670

alexey-kravtsov:


src/main/java/com/axibase/tsd/api/model/series/SeriesQuery.java
```diff
@@ -23,7 +23,7 @@
     private String startDate;
     private String endDate;
     private Interval interval;
-    private Map<String, String> tags = new HashMap<>();
```
vtols: Is there a special reason for such a change?

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
```diff
@@ -551,6 +548,93 @@ public void testLastSeriesWithText() throws Exception {
...
+    @Issue("4670")
+    @Test(
+            dataProvider = "provideTagFilters",
+            description = "test series query with tag filter")
+    public void testTagFilter(Filter<Series> filter) {
+        String filterExpression = StringUtils.isEmpty(filter.getExpression())
+                ? ""
+                : "," + filter.getExpression();
+
+        String payload = String.format("{ " +
```
vtols: Why is it problematic to use SeriesQuery here?

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
```diff
@@ -551,6 +548,93 @@ public void testLastSeriesWithText() throws Exception {
...
+                "\"entity\": \"%s\"," +
+                "\"metric\": \"%s\"" +
+                "%s" +
+                "}",
+                TEST_SERIES1.getEntity(),
+                TEST_SERIES1.getMetric(),
+                filterExpression);
+
+        Response response = SeriesMethod.querySeries(payload);
+        List<Series> result = response.readEntity(new GenericType<List<Series>>() {});
```
vtols: Is it possible to simply use `Series[].class` here instead of `new GenericType<List<Series>>() {}`?

src/main/java/com/axibase/tsd/api/model/series/SeriesQuery.java
alexey-kravtsov: Because when tags aren't set, empty tag field is send. Now no tags field sent when no tags specified

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
alexey-kravtsov: Because it is impossible to send "tags: null"

rmakulov:
Please, resolve conflicts.

PR #446, opened 2017-11-16
Data API: processing error in series query method with different periods #4714

alexey-kravtsov:


MikhailZvagelsky:
Bug #4714#note-10 fixed.

PR #450, opened 2017-11-22
Test for #4713

vtols:


src/test/java/com/axibase/tsd/api/method/series/SeriesQueryAddMetaTest.java
```diff
@@ -0,0 +1,72 @@
...
+        return result;
+    }
+
+    @Issue("4713")
+    @Test(
+            description = "Check that meta is included for all types of data",
+            dataProvider = "seriesQueryTypeProvider"
+    )
+    public void testSeriesResponseMetaIncluded(SeriesQueryType type) throws JSONException {
+        long currentTime = System.currentTimeMillis();
```
rmakulov: Will the tests remain reproducible? It seems to me that forecasts can be made also for the series in past.

rmakulov: How do you build forecasts - do we need to update api_test image?

vtols: There is default forecast, I think we don't need to create a new one specially for this test

vtols: But if we want to create forecasts in the past, then possibly we need to create a new forecasting rule

PR #451, opened 2017-11-22
Tests for #4715

vtols:


src/test/java/com/axibase/tsd/api/method/metric/MetricSeriesTagsTest.java
```diff
@@ -0,0 +1,152 @@
...
+
+    @Issue("4715")
+    @Test(description = "Test {metric}/series/tags with single tags parameter")
+    public void testMetricSeriesTagsTagParam() throws Exception {
+        String expectedJson = "{" +
+                "  \"t1\" : [ \"p1\", \"p2\", \"v1\", \"v2\", \"x1\" ]" +
+                "}";
+
+        String r = queryMetricSeriesTags(METRIC_NAME,
+                new MetricListParameters().addTag("t1")).readEntity(String.class);
```
rmakulov: Please, extract second parameter into variable here and further to improve readability.
For what r stands for?)

vtols: Sorry, just forgot to give it name

src/main/java/com/axibase/tsd/api/method/metric/MetricMethod.java
```diff
@@ -96,6 +98,18 @@ public static Response deleteMetric(String metricName) throws Exception {
         return response;
     }
 
+    public static Response queryMetricSeriesTags(String metricName,
```
alexey-kravtsov: I suppose it would be easier to add method that accepts class (MetricSeriesTagsQuery) and returns class (MetricSeriesTags) and use this method in tests instead of querying and comparing json. This method may be used to test for errors

PR #454, opened 2017-12-11
Test for #4756

vtols:


src/main/java/com/axibase/tsd/api/model/series/Sample.java
```diff
@@ -36,6 +36,11 @@
     @JsonProperty("v")
     private BigDecimal value;
 
+    @JsonDeserialize(using = ValueDeserializer.class)
+    @JsonTypeInfo(use = JsonTypeInfo.Id.NONE)
```
vtols: I don't know why do we need this annotation. Maybe we can remove it.

PR #456, opened 2017-12-18
Tests and updates for #4803

vtols:


src/test/java/com/axibase/tsd/api/method/sql/function/interpolate/InterpolationFillModeTest.java
```diff
@@ -0,0 +1,269 @@
...
+    @Test(description = "Check that interpolation work with fill mode VALUE {n}," +
+            "WITH INTERPOLATE(... VALUE {n} ...)")
+    public void testInterpolationFillValueMode() {
+        String sqlQuery = String.format(
+                "SELECT datetime, value FROM \"%s\" " +
+                        "WHERE datetime BETWEEN '2017-12-15T11:00:00.000Z' AND '2017-12-15T15:00:00.000Z' " +
+                        "WITH INTERPOLATE(%s) " +
+                        "ORDER BY datetime",
+                METRIC_NAME,
+                new InterpolationParams(30, TimeUnit.MINUTE).linear().fill(-1.0).timeZone("UTC")
```
alexey-kravtsov: I suppose it would be simpler and more clear to write WITH INTERPOLATE(30 MINUTE, LINEAR, VALUE -1, 'UTC') instead of using this constructor

src/main/java/com/axibase/tsd/api/model/sql/function/interpolate/FillMode.java
```diff
@@ -1,5 +1,49 @@
...
+            mode = "NAN";
+        } else {
+            mode = "VALUE " + String.valueOf(val);
+        }
+        return new FillMode(mode, true);
+    }
+
+    public static List<String> stringValues() {
+        List<String> result = new ArrayList<>();
+        result.addAll(STRING_VALUE_SET);
```
rmakulov: List<String> result = new ArrayList<>(STRING_VALUE_SET);

src/test/java/com/axibase/tsd/api/method/sql/function/interpolate/InterpolationFillModeTest.java
```diff
@@ -0,0 +1,269 @@
...
+        Series series = new Series(ENTITY_NAME, METRIC_NAME);
+        series.addSamples(Sample.ofDateInteger("2017-12-15T12:00:00.000Z", 1));
+        series.addSamples(Sample.ofDateInteger("2017-12-15T13:00:00.000Z", 2));
+        series.addSamples(Sample.ofDateInteger("2017-12-15T14:00:00.000Z", 3));
+
+        SeriesMethod.insertSeriesCheck(series);
+    }
+
+    @Issue("4803")
+    @Test(description = "Check that interpolation work with fill mode VALUE {n}," +
```
rmakulov: *works?

src/test/java/com/axibase/tsd/api/method/sql/function/interpolate/InterpolationFillModeTest.java
vtols: I use it to avoid writing all of default arguments before timezone argument explicitly

PR #457, opened 2017-12-20
Fixed PREVIOUS interpolation tests

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/function/interpolate/InterpolationBoundaryValuesTest.java
```diff
@@ -67,187 +45,85 @@ public void prepareData() throws Exception {
         SeriesMethod.insertSeriesCheck(series1, series2);
     }
 
-    private String[][] generateCalendarInterpolationOutput(Boundary boundary, DateRange... ranges) {
```
vtols: +1 for removing server timezone-dependence :smiley:

PR #462, opened 2018-01-15
SQL: Incorrect GROUP BY PERIOD END_TIME with DST change inside period #4700

alexey-kravtsov:


src/test/java/com/axibase/tsd/api/method/sql/function/interpolate/InterpolationBoundaryValuesTest.java
```diff
@@ -364,7 +364,7 @@ public void testOuterInterpolationWithWithOuterBoundValueSinglePeriod() {
     }
 
     @Issue("4217")
-    @Test(enabled = false)
+    @Test
```
vtols: :+1:

PR #463, opened 2018-01-18
Data API: END_TIME aggregation #4867, refactoring

alexey-kravtsov:


src/main/java/com/axibase/tsd/api/model/message/MessageStatsQuery.java
```diff
@@ -1,7 +1,7 @@
 package com.axibase.tsd.api.model.message;
 
 import com.axibase.tsd.api.model.Period;
-import com.axibase.tsd.api.model.series.Aggregate;
+import com.axibase.tsd.api.model.series.query.transformation.aggregate.Aggregate;
```
vtols: Why so long?

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryTest.java
```diff
@@ -847,15 +856,109 @@ public void testDoubleSeriesQueryDifferentAggregationPeriodAlign() throws Except
...
+        query.setAggregate(aggregate);
+
+        List<Series> result = SeriesMethod.querySeriesAsList(query);
+
+        Series expectedSeries = new Series();
+        expectedSeries.setEntity(TEST_SERIES3.getEntity());
+        expectedSeries.setMetric(TEST_SERIES3.getMetric());
+        expectedSeries.setTags(Mocks.TAGS);
+        expectedSeries.addSamples(
+                Sample.ofDateDecimal("2017-01-01T00:01:30Z", new BigDecimal("2.0")),
```
vtols: Why do you need BigDecimal here

alexey-kravtsov: Because there are no double constructor methods for Sample. They were not added due to double-to-string rounding issues. Some examples here https://stackoverflow.com/questions/24776587/converting-a-double-to-exact-string-in-java

src/main/java/com/axibase/tsd/api/model/message/MessageStatsQuery.java
alexey-kravtsov: According to documentation https://github.com/axibase/atsd/blob/master/api/data/series/query.md#transformation-fields , Aggregate is a transformation function. This is the reason why it is in transformation package. Do you suggest more simple/short names?

src/main/java/com/axibase/tsd/api/model/series/query/transformation/interpolate/InterpolateFunction.java
```diff
@@ -0,0 +1,7 @@
+package com.axibase.tsd.api.model.series.query.transformation.interpolate;
+
+public enum InterpolateFunction {
+    LINEAR,
+    PREVIOUS,
+    AUTO
+}
```
rmakulov: Do we need both? https://github.com/axibase/atsd-api-test/blob/316e437b03ceebc862b42ea096ebf56f3e73a06e/src/main/java/com/axibase/tsd/api/model/sql/function/interpolate/InterpolateFunction.java#L4

alexey-kravtsov: I suppose yes, because first is specific for API interpolation, and second for SQL interpoaltaion. They may be changed independently, like interpolation option for GROUP BY
API: https://github.com/axibase/atsd/blob/master/api/data/series/group.md#interpolation-functions
SQL: https://github.com/axibase/atsd/tree/master/sql#interpolation
Note that SQL hasn't NONE field

src/test/java/com/axibase/tsd/api/method/series/SeriesQueryGroupExampleTest.java
```diff
@@ -94,7 +101,7 @@ public void testExampleSumInterpolation() throws Exception {
     }
 
     @Issue("2995")
-    @Test(description = "https://github.com/axibase/atsd-docs/blob/master/api/data/series/group.md#group-aggregation")
+    @Test(description = "https://github.com/axibase/atsd-docs/blob/master/api/data/series/group.md#group-transformation")
```
rmakulov: Please, check the links.

PR #480, opened 2018-03-26
SQL: Сolumn name sensitivity in subqueries #5139

unrealwork:
### Added additional test cases

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/SqlTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/SqlTest.java
@@ -120,9 +120,9 @@ public abstract class SqlTest extends SqlMethod {
         // If expected result is not empty, but actual is empty, wait 100ms and try again
         // See #5057
         if (expectedRows.size() > 0) {
-            for (int i = 0; i<10 &&  resultTable.getRows().size() == 0; i++) {
+            for (int timeout = 100; timeout <= 1600 && resultTable.getRows().size() == 0; timeout *= 2) {
                 try {
-                    Thread.sleep(200);
+                    Thread.sleep(timeout);
                 } catch (InterruptedException e) {
                     e.printStackTrace();
                 }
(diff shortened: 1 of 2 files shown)


src/test/java/com/axibase/tsd/api/method/sql/SqlTest.java
```diff
@@ -116,17 +116,18 @@ public void assertRowsMatch(String message, String[][] expectedRows, StringTable
...
-
         // Some series may be not returned immediately after insert.
         // If expected result is not empty, but actual is empty, wait 100ms and try again
         // See #5057
-        if (expectedRows.size() > 0 && resultTable.getRows().size() == 0) {
-            try {
-                Thread.sleep(100);
-            } catch (InterruptedException e) {
-                e.printStackTrace();
+        if (expectedRows.size() > 0) {
```
rodionos: I don't feel good about this because the question remains: why is data not instantly visible in SQL?
But we're going to implement a hack, lets do an exponential backoff:
timeouts = 0, 100, 200, 400, 800, 1600

src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlSubqueryTest.java
```diff
@@ -368,4 +373,159 @@ public void testTimeExpression() {
...
+        };
+
+        assertSqlQueryRows("Wrong result when using lower-cased column aliases in subquery", expectedRows, sqlQuery);
+    }
+
+    @Issue("5139")
+    @Test(description = "Test mixed-case column aliases in subquery",
+            dataProvider = "sqlSubqueryMixedCaseDataProvider")
+    public void testMixedCaseAlias(final AllTypeAliases innerAliases, final AllTypeAliases outerAliases) {
+        String sqlQuery = MessageFormat.format(
```
vtols: Why do you use both `MessageFormat.format()` and `String.format()`?

unrealwork: There are too much format arguments, identifiers help to understand where is position of an  argument in formatting String.

src/test/java/com/axibase/tsd/api/method/sql/SqlTest.java
```diff
@@ -116,17 +116,18 @@ public void assertRowsMatch(String message, String[][] expectedRows, StringTable
 
     public void assertSqlQueryRows(String message, List<List<String>> expectedRows, String sqlQuery) {
         StringTable resultTable = queryTable(sqlQuery);
```
unrealwork: @rodionos There is request without a timeout, then if there were not enough expected rows we are trying to sleep.

PR #488, opened 2018-04-19
revert debug changes

raipc:


raipc:
It seems like buffering request entities fixes the problem with flaky tests. I reverted the code handling ClientProtocolException.

PR #498, opened 2018-06-15
#5389 Use new Replacement Table meta API in ReplacementTableMethod

kirmanak:


rodionos:
@unrealwork @raipc - please review the replacement table refactoring completed by @kirmanak

README.md
```diff
@@ -1,7 +1,100 @@
...
 
+ATSD configuration<a name="rule"></a>
+==================
 The API tests require the target ATSD to be properly configured:
 
 1. Set `last.insert.write.period.seconds` to 0 on the **Settings > Server Properties** page.
 
 2. Import and enable the [test rule](https://raw.githubusercontent.com/axibase/dockers/atsd_api_test/rules.xml) on the **Alerts > Rules** page.
+
+API tests re-run
```
raipc: We have a separate image for tests. Create a container for each test execution. There is no need to update ATSD, modify server properties, or import rules.

```sh
docker run -d --name="atsd-api-test" -e axiname="$ATSD_LOGIN" -e axipass="$ATSD_PASSWORD" -e timezone="Asia/Kathmandu" axibase/atsd:api_test
```

README.md
```diff
@@ -1,7 +1,100 @@
 [![Build Status](https://travis-ci.org/axibase/atsd-api-test.svg?branch=master)](https://travis-ci.org/axibase/atsd-api-test)
 
+ATSD configuration<a name="rule"></a>
```
rodionos: Use vscode editor. Install mardownlist and spellcheck extensions to fix formatting problems. Our rules are: https://github.com/axibase/atsd/blob/master/.markdownlint.json

kirmanak: This is not a solution to a formatting problem, this is an anchor for referencing this chapter from the next.

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -50,58 +33,42 @@ private static Response createResponse(ReplacementTable table) {
...
-        Element table = doc.select("table").get(0);
-        Elements trs = table.select("tr");
-        for (Element tr : trs) {
-            Elements tds = tr.select("td");
-            for (Element td : tds) {
-                if (td.text().contains(replacementTable)) {
-                    return true;
-                }
-            }
+        if (replacementTable.getName().equals(replacementTableName)) {
```
raipc: I think that `equalsIgnoreCase` method should be used because replacement table names are case-insensitive. Also I would put this check inside previous try-catch block for simplicity.

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -3,44 +3,27 @@
...
-import java.util.Map;
 
+import static javax.ws.rs.core.Response.Status.NOT_FOUND;
 import static javax.ws.rs.core.Response.Status.OK;
 
 public class ReplacementTableMethod extends BaseMethod {
     private static final Logger logger = LoggerFactory.getLogger(ReplacementTableMethod.class);
-    private static final String METHOD_REPLACEMENT_TABLE_NEW = "/replacement-tables/new";
-    private static final String METHOD_REPLACEMENT_TABLE_LIST = "/replacement-tables/";
+    private static final String API_V1_REPLACEMENT_TABLES_JSON = "/api/v1/replacement-tables/json/";
```
unrealwork: It's unnecessary to use `/api/v1` prefix in api path, execute you request through `executeApiRequest` method from `BaseMethod` class.

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -3,44 +3,27 @@
...
-        MultivaluedMap<String, String> parameters = new MultivaluedHashMap<>();
-        parameters.add("lookupName", table.getName());
-        parameters.add("items", squashMapIntoString(table.getMap()));
-        parameters.add("oldName", "");
-        parameters.add("save", "Save");
-        Form form = new Form(parameters);
-
         Response response = executeRootRequest(webTarget -> webTarget
-                .path(METHOD_REPLACEMENT_TABLE_NEW)
+                .path(API_V1_REPLACEMENT_TABLES_JSON + table.getName())
```
unrealwork: Use template path instead concatenation and resolve it with [`resolveTemplate`](https://docs.oracle.com/javaee/7/api/javax/ws/rs/client/WebTarget.html#resolveTemplate-java.lang.String-java.lang.Object-) method from `WebTarget` class. [Example](https://github.com/axibase/atsd-api-test/blob/master/src/main/java/com/axibase/tsd/api/method/entity/EntityMethod.java#L28) .

unrealwork:
@kirmanak Where are methods to get List of replacement tables names and delete specified replacement table? [Docs](https://github.com/axibase/atsd/blob/096d6eb6a6f4f711f531724418314b19af8c6dcf/api/meta/replacement-table/README.md)

kirmanak:
@unrealwork  Since there wasn't such methods earlier and I was asked to replace old parsing with new API, I didn't write any new methods. Should I do it?

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -50,58 +33,40 @@ private static Response createResponse(ReplacementTable table) {
...
+                .path(API_V1_REPLACEMENT_TABLES_JSON + replacementTableName)
                 .request().get());
         response.bufferEntity();
         return response;
     }
 
-    public static boolean replacementTableExist(String replacementTable) throws NotCheckedException {
-        replacementTable = replacementTable.replace(" ", "_").toLowerCase();
-        final Response response = ReplacementTableMethod.getReplacementTablesResponse();
+    public static boolean replacementTableExist(String replacementTableName) throws NotCheckedException {
```
unrealwork: I think, it will be easier to check that list of replacement-tables names contains specified name. And it will be better to implement method that checks `ReplacementTable` instance existing.

src/main/java/com/axibase/tsd/api/model/replacementtable/ReplacementTable.java
```diff
@@ -7,13 +7,13 @@
 
 public class ReplacementTable {
```
unrealwork: Add all fields available in new endpoint [response](https://github.com/axibase/atsd/blob/096d6eb6a6f4f711f531724418314b19af8c6dcf/api/meta/replacement-table/get.md#response).

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
unrealwork: @raipc Easier to implement, just check `getList().contains(formattedName)`?

unrealwork:
@kirmanak Just replace existing methods, I think new methods will be implemented in another task. Sorry, I will remove unnecessary comments.

kirmanak:
@unrealwork I guess, changes in ReplacementTable were necessary, so I made new fields, getters, setters, constructors.

src/main/java/com/axibase/tsd/api/model/replacementtable/ReplacementTable.java
```diff
@@ -6,30 +6,77 @@
...
     }
 
+    public ReplacementTable(String name, String author, String description) {
+        Registry.ReplacementTable.checkExists(name);
+        this.name = name;
+        this.author = author;
+        this.description = description;
+    }
+
+    public ReplacementTable(String name, String description, String author, SupportedFormat format, Map<String, String> keys) {
```
raipc: replace getters, setters, all-args constructor with Lombok

kirmanak: Lombok will not generate the check in Registry

raipc:
Please, remove `Registry.ReplacementTable.checkExists(name);` from constructors and remove unused constructors

src/main/java/com/axibase/tsd/api/model/replacementtable/ReplacementTable.java
```diff
@@ -1,35 +1,25 @@
...
-        this.name = name;
-        this.map = map;
-    }
+    private String name;
+    private String description;
+    private String author;
+    private SupportedFormat format;
+    private Map<String, String> keys = new HashMap<>();
 
     public ReplacementTable(String name) {
```
raipc: Replace the constructor with static factory method `of(String name)`. It's dangerous to call `checkExists` from the constructor : it may lead to infinite loops.

unrealwork: I think `format` field also is required to add to the constructor.

kirmanak: You asked me to delete the call ``checkExists``, but now you're asking me to replace constructor with factory method because of ``checkExists``. The constructor is used by ``Response.readEntity()`` in ``ReplacementTableMethod.java``, so we can't just remove it.

raipc: Yes, the old behavior should stay (`checkExists` called), but constructor should not be used for it. Constructor may be called by Jackson in checkExists method

README.md
```diff
@@ -1,7 +1,8 @@
 [![Build Status](https://travis-ci.org/axibase/atsd-api-test.svg?branch=master)](https://travis-ci.org/axibase/atsd-api-test)
 
-The API tests require the target ATSD to be properly configured:
+Run tests in a freshly installed Docker container with image for tests:
 
-1. Set `last.insert.write.period.seconds` to 0 on the **Settings > Server Properties** page.
```
rodionos: Keep the option of configuring a regular ATSD instance listed.
Using a special image can be listed as another option.

rodionos: Add information on how to delete all tables from an existing ATSD instance.

README.md
```diff
@@ -1,7 +1,82 @@
 [![Build Status](https://travis-ci.org/axibase/atsd-api-test.svg?branch=master)](https://travis-ci.org/axibase/atsd-api-test)
 
-The API tests require the target ATSD to be properly configured:
+How to run tests
+================
```
rodionos: This is a minor tweak, but important for consistency. Use syntax constructs from the [github markdown guide](https://guides.github.com/features/mastering-markdown/). Specifically, use hashes for section headings:

```markdown
# Running Tests
```

Our [guidelines](https://github.com/axibase/docs-util/blob/master/guide.md) apply as well, which require **title case** for headings and backticked parameter names and values, among other things.

src/main/java/com/axibase/tsd/api/model/replacementtable/ReplacementTable.java
```diff
@@ -1,35 +1,32 @@
...
+    private String name;
+    private String description;
+    private String author;
+    private SupportedFormat format;
+    private Map<String, String> keys = new HashMap<>();
 
-
-    public ReplacementTable(String name, Map<String, String> map) {
-        Registry.ReplacementTable.checkExists(name);
+    private ReplacementTable(String name, SupportedFormat format) {
```
raipc: If the class is instantiated using reflection (as in `response.readEntity(Class c)`), there must be a default constructor. I suggest to add it using lombok annotation `@NoArgsConstructor`. Defined private constructor suppressed the compiler to generate a default constructor. In this case `ProcessingException` was thrown in `response.readEntity` which was then wrapped into `NotCheckedException` which means that the replacement table doesn't exist (even if it does).

Lombok annotation `@Accessors(chain = true)` can be useful here to set field values.

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -50,58 +34,42 @@ private static Response createResponse(ReplacementTable table) {
...
+                .resolveTemplate("table", replacementTableName)
                 .request().get());
         response.bufferEntity();
         return response;
     }
 
-    public static boolean replacementTableExist(String replacementTable) throws NotCheckedException {
-        replacementTable = replacementTable.replace(" ", "_").toLowerCase();
-        final Response response = ReplacementTableMethod.getReplacementTablesResponse();
+    public static boolean replacementTableExist(String replacementTableName) throws NotCheckedException {
```
unrealwork: `NotCheckedException` is RuntimeException. It shouldn't be declared in method specification.

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -50,58 +34,42 @@ private static Response createResponse(ReplacementTable table) {
...
-        Element table = doc.select("table").get(0);
-        Elements trs = table.select("tr");
-        for (Element tr : trs) {
-            Elements tds = tr.select("td");
-            for (Element td : tds) {
-                if (td.text().contains(replacementTable)) {
-                    return true;
-                }
+            ReplacementTable replacementTable = response.readEntity(ReplacementTable.class);
+            if (!replacementTable.getName().equalsIgnoreCase(replacementTableName)) {
```
unrealwork: There is potential `NPE`, better to compare it using [StringUtils.equalsIgnoreCase](https://commons.apache.org/proper/commons-lang/javadocs/api-2.6/org/apache/commons/lang/StringUtils.html#equalsIgnoreCase(java.lang.String,%20java.lang.String))

src/main/java/com/axibase/tsd/api/model/replacementtable/ReplacementTable.java
```diff
@@ -1,35 +1,35 @@
...
-    public String getName() {
-        return name;
-    }
-
-    public Map<String, String> getMap() {
-        return map;
+        return new ReplacementTable(name, format);
     }
 
     public void addValue(String key, String value){
```
unrealwork: May be make the method also with chain access? It will  reduce table creation.

```java
Replacement.of(name, SupportedFormat.LIST)
        .addValue("-1", "negative")
        .addValue("1", "positive")
        .addValue("2", "2")
        .addValue("word", "3");
```

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
```diff
@@ -50,58 +38,46 @@ private static Response createResponse(ReplacementTable table) {
...
         response.bufferEntity();
         return response;
     }
 
-    public static boolean replacementTableExist(String replacementTable) throws NotCheckedException {
-        replacementTable = replacementTable.replace(" ", "_").toLowerCase();
-        final Response response = ReplacementTableMethod.getReplacementTablesResponse();
+    public static boolean replacementTableExist(String replacementTableName) {
+        replacementTableName = replacementTableName.replace(" ", "_").toLowerCase();
+        final Response response = ReplacementTableMethod.getReplacementTablesResponse(replacementTableName);
```
unrealwork: `ReplacementTableMethod.` is not necessary inside own methods.

src/main/java/com/axibase/tsd/api/method/replacementtable/ReplacementTableMethod.java
raipc: I think that RuntimeException must be declared if it is used for control flow. The discussion of disadvantages of this approach is out of scope of this pull request.

unrealwork:
@rodionos Should `README.md` have reference to our [contributing](https://github.com/axibase/atsd-api-test/blob/master/CONTRIBUTING.md) guide?

rodionos:
contrib will stay separate for now, merging

PR #506, opened 2018-07-30
5494: Add a big query to test WHERE and HAVING is_weekday()

kirmanak:


src/test/java/com/axibase/tsd/api/method/sql/clause/where/SqlWhereIsWeekdayTest.java
```diff
@@ -0,0 +1,66 @@
...
+                .addSamples(Sample.ofDateInteger("2018-01-06T00:00:00Z", 6))
+                .addSamples(Sample.ofDateInteger("2018-01-07T00:00:00Z", 7))
+                .addSamples(Sample.ofDateInteger("2018-03-02T00:00:00Z", 2))
+                .addSamples(Sample.ofDateInteger("2018-03-03T00:00:00Z", 3))
+                .addSamples(Sample.ofDateInteger("2018-03-04T00:00:00Z", 4));
+
+        SeriesMethod.insertSeriesCheck(seriesFirst, seriesSecond);
+    }
+
+    @Test(description = "Test WHERE and HAVING with every possible clause in one query")
```
unrealwork: Are you testing all clauses or `WHERE` and `HAVING`? Split into more specific test cases to localize an error. Assert message should describe where and how an error happened and which data caused the error.

kirmanak: @unrealwork Sergei asked me to add this particular query to API tests, I am not sure should I modify it or not.

rodionos: Splitting would be better so that the functions use in each is tested separately.

kirmanak: @rodionos I've added several smaller tests, but saved the big one. If the big one fails because of these functions (`IS_WEEKDAY`, `IS_WORKDAY`), at least one of the smaller will fail too, I guess. This way we can test the function in a big query and quickly understand which part caused an error. Is this solution ok?

src/test/java/com/axibase/tsd/api/method/sql/clause/where/SqlWhereIsWeekdayTest.java
```diff
@@ -0,0 +1,157 @@
...
+
+    @Issue("5494")
+    @Test(
+            description = "Test the functions in SELECT clause",
+            dataProvider = "provideSelectQueries"
+    )
+    public void testSelect(final String params, final String[] results) {
+        final String query = String.format("SELECT %s FROM \"%s\"", params, METRIC_NAME_1);
+        final String[][] expectedRows = expectedRows(results);
+        assertSqlQueryRows("Fail to use boolean function after SELECT keyword", expectedRows, query);
```
unrealwork: Which boolean function? Add Test case params to assert message in all tests.

src/test/java/com/axibase/tsd/api/method/sql/clause/where/SqlWhereIsWeekdayTest.java
```diff
@@ -0,0 +1,157 @@
...
+                        ")\n" +
+                        "group by period(1 day, PREVIOUS, EXTEND)\n" +
+                        "with time >= last_time - 1*MONTH\n" +
+                        "order by is_weekday(datetime, 'ISR')\n" +
+                        "limit 10\n", METRIC_NAME_1, METRIC_NAME_2);
+        final String[][] expectedRows = toArray(
+                toArray("false", "true", "true", "false"),
+                toArray("false", "true", "true", "false")
+        );
+        assertSqlQueryRows("Fail to execute a big query (anything could go wrong)", expectedRows, query);
```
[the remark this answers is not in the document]
unrealwork: Yes, I've just described its behavior.

PR #507, opened 2018-08-06
5538: Fix testCanSetEmptyExpression()

kirmanak:
Tests are failing because of `javax.ws.rs.client.Entity#json` method, which does not send empty fields of the object. So, when we set `expression` field to an empty string, we don't send it to the server. According to [docs](https://github.com/axibase/atsd/blob/master/api/meta/entity-group/update.md) unspecified fields are left unchanged, that's why the test fails. 
I can't change `javax.ws.rs.client.Entity#json` method, I can't replace it. I tried to send a space character to force `javax.ws.rs.client.Entity#json` method send this field, but the ATSD trims the field before the storing, so the response for the GET method later does not contain `expression` field (it is empty after the trimming, `javax.ws.rs.client.Entity#json` doesn't send it) and the test fails. 
That's why my solution is so ugly, though it works.

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/entitygroup/EntityGroupUpdateTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/entitygroup/EntityGroupUpdateTest.java
@@ -44,9 +44,10 @@ public class EntityGroupUpdateTest extends EntityGroupMethod {
         entityGroup.setExpression(SYNTAX_ALLOWED_ENTITYGROUP_EXPRESSION);
         createOrReplaceEntityGroupCheck(entityGroup);
 
-        entityGroup.setExpression("");
+        entityGroup.setExpression(" ");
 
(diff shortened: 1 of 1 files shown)


rodionos:
@raipc @unrealwork - please check what is wrong with our API and why hacks are required. We're just fixing a bug that was present in 5538.

unrealwork:
@rodionos The test was incorrect, our model of EntityGroup didn't allow to send empty expression (as result of @JsonInclude(JsonInclude.Include.NON_EMPTY) applied to `expresion` field). Therefore we sent path request with not specified expression: 

```json
 > PATCH  http://localhost:50088/api/v1/entity-groups/update-entitygroup-4
{
  "name" : "update-entitygroup-4",
  "tags" : { },
  "enabled" : true
}

```

As mention in 5538  the update method was incorrect and the request changed the expression to empty .

PR #511, opened 2018-08-10
5542: Add WITH TIMEZONE tests

kirmanak:
I know, these tests are quite simple.
I just wanted to show, that work in progress and let anybody express the opinion.

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
@@ -65,8 +65,8 @@ public class SqlWithTimezoneTest extends SqlTest {
         Object[][] results = null;
         for (final ZoneId timeZone : timeZones) {
             results = ArrayUtils.addAll(results, toArray(
-                    testCase("date_format(time, 'yyyy-MM-ddTHH:mm:ss')",
-                            timeZone, getTimeStream(timeZone)
+                    testCase("date_format(time, 'yyyy-MM-ddTHH:mm:ss')", timeZone,
+                            getTimeStream(timeZone)
                                     .map((time) -> time.format(ISO_LOCAL_DATE_TIME))
                     ), testCase("date_format(dateadd(minute, -15, time), 'yyyy-MM-ddTHH:mm:ss')", timeZone,
                             getTimeStream(timeZone)
@@ -202,17 +202,6 @@ public class SqlWithTimezoneTest extends SqlTest {
         return results;
     }
 
-    private static String listToString(final List<String> list) {
-        final StringBuilder builder = new StringBuilder();
-        final ListIterator<String> iterator = list.listIterator();
-        while (iterator.hasNext()) {
-            builder.append(iterator.next());
-            if (iterator.hasNext()) {
-                builder.append(", ");
-            }
-        }
-        return builder.toString();
-    }
 
     @DataProvider
     public static Object[][] provideInterpolatedData() {
@@ -307,7 +296,7 @@ public class SqlWithTimezoneTest extends SqlTest {
                 "WHERE %s WITH TIMEZONE = \'%s\'", METRIC_NAME, expression, timeZone
         );
         final String assertMessage =
-                String.format("Failed to modify timezone via WITH TIMEZONE in WHERE \"%s\"", expression);
+                String.format("Failed to filter rows using WITH TIMEZONE in WHERE \"%s\"", expression);
         assertSqlQueryRows(assertMessage, expectedRows, condition);
     }
 
@@ -321,7 +310,7 @@ public class SqlWithTimezoneTest extends SqlTest {
                 expression, METRIC_NAME, timeZone
         );
         final String assertMessage =
-                String.format("Failed to modify timezone via WITH TIMEZONE in SELECT \"%s\"", expression);
+                String.format("Failed to query rows using WITH TIMEZONE in SELECT \"%s\"", expression);
         assertSqlQueryRows(assertMessage, expectedRows, query);
     }
 
@@ -334,6 +323,6 @@ public class SqlWithTimezoneTest extends SqlTest {
         final String query = String.format("SELECT SUM(value) FROM \"%s\" GROUP BY PERIOD(%s) WITH TIMEZONE = \"%s\"",
                 METRIC_NAME, period, timeZone
         );
-        assertSqlQueryRows("Fail to group by date with timezone", expectedRows, query);
+        assertSqlQueryRows("Fail to group by date using WITH TIMEZONE", expectedRows, query);
     }
 }
--- a/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
@@ -35,6 +35,7 @@ import static java.util.stream.Collectors.summingInt;
 import static org.apache.commons.lang3.ArrayUtils.toArray;
 
 public class SqlWithTimezoneTest extends SqlTest {
+    private static int flakyRepeatCounter = 1;
     private static final String ENTITY_NAME = entity();
     private static final String METRIC_NAME = metric();
     private static final ZoneId[] timeZones = toArray(
@@ -94,7 +95,19 @@ public class SqlWithTimezoneTest extends SqlTest {
                     ), testCase("date_format(date_parse('2018-02-02T15:30:00', 'yyyy-MM-ddTHH:mm:ss'), " +
                                     "'yyyy-MM-ddTHH:mm:ss')", timeZone,
                             SAMPLES.stream().map((sample) -> "2018-02-02T15:30:00")
-                    ), testCase("current_day", timeZone,
+                    )
+            ));
+        }
+
+        return results;
+    }
+
+    @DataProvider
+    public static Object[][] provideCalendarKeywords() {
+        Object[][] results = null;
+        for (final ZoneId timeZone : timeZones) {
+            results = ArrayUtils.addAll(results, toArray(
+                    testCase("current_day", timeZone,
                             SAMPLES.stream().map((sample) -> ZonedDateTime.now(timeZone))
                                     .map((zoned) -> zoned.with(SECOND_OF_DAY, 0))
                                     .map(ChronoZonedDateTime::toEpochSecond)
@@ -218,6 +231,18 @@ public class SqlWithTimezoneTest extends SqlTest {
         }).toArray(Object[][]::new);
     }
 
+    private static String listToString(final List<String> list) {
+        final StringBuilder builder = new StringBuilder();
+        final ListIterator<String> iterator = list.listIterator();
+        while (iterator.hasNext()) {
+            builder.append(iterator.next());
+            if (iterator.hasNext()) {
+                builder.append(", ");
+            }
+        }
+        return builder.toString();
+    }
+
     @Issue("5542")
     @Test(
             dataProvider = "provideInterpolatedData",
@@ -243,6 +268,33 @@ public class SqlWithTimezoneTest extends SqlTest {
 
     private static boolean isWeekday(final ZonedDateTime dateTime) {
         return !dateTime.getDayOfWeek().equals(DayOfWeek.SATURDAY) && !dateTime.getDayOfWeek().equals(DayOfWeek.SUNDAY);
+    }
+
+    @Issue("5542")
+    @Test(
+            dataProvider = "provideCalendarKeywords"
+    )
+    public void testCalendarKeywords(final String expression, final String timeZone, final String[][] expectedRows) {
+        final String query = String.format("SELECT %s FROM \"%s\" WITH TIMEZONE = \'%s\'",
+                expression, METRIC_NAME, timeZone
+        );
+        final String assertMessage =
+                String.format("Failed to modify timezone via WITH TIMEZONE in SELECT \"%s\"", expression);
+        try {
+            assertSqlQueryRows(assertMessage, expectedRows, query);
+        } catch (final AssertionError err) {
+            if (flakyRepeatCounter == 0) {
+                throw err;
+            }
+            flakyRepeatCounter--;
+            final Object[][] parameters = provideCalendarKeywords();
+            for (final Object[] arguments : parameters) {
+                final String expr = (String) arguments[0];
+                final String tz = (String) arguments[1];
+                final String[][] expected = (String[][]) arguments[2];
+                testCalendarKeywords(expr, tz, expected);
+            }
+        }
     }
 
     @Issue("5542")
--- a/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
@@ -15,7 +15,6 @@ import java.time.DayOfWeek;
 import java.time.LocalDateTime;
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
-import java.time.chrono.ChronoZonedDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


rodionos:
Search https://axibase.com/docs/atsd/sql/ for syntax options that either always use database time zone or allow for custom database time zone. Suggested keywords:

`time zone`, `timezone`, `database time`, `server time`.

Enumerate functions and clauses which TZ-dependent, for example, https://axibase.com/docs/atsd/sql/#interval-condition or https://axibase.com/docs/atsd/sql/#period which support dates in server time zone as well as calendar keywords which are TZ-dependent, e.g.:

```sql
BETWEEN '2018' AND '2018-02-01'
datetime >= '2018-02-01'
datetime >= CURRENT_MONTH
PERIOD (1 DAY)
```

For each function/clause/condition add test to check that the date is parsed and processed as expected. For example, to test PERIOD, extract timezone from ATSD, calculate PERIOD start date in this time zone, then execute a query that performs GROUP BY using WITH timezone and compare the returned and calculated PERIOD start dates.

kirmanak:
@rodionos Why do you want me to use ATSD timezone? Calculating test results is not as reliable as hardcoding them, I guess. I offer this approach: insert data in `UTC` timezone, perform requests in `Asia/Kathmandu` and then the same requests in `Pacific/Marquesas`, compare results.

rodionos:
How data is inserted is not important because internally it is stored in the database without timezone information. Once data is inserted, I'd like the tests to return correct results in any timezone. In previous versions we used other timezones, not just`Asia/Kathmandu` `Pacific/Marquesas`. We might use different timezone tomorrow. Hence my suggestion is to perform queries in the same timezone as the currently returned by the database. You can hardcode expected results for all possible 200+ timezones or you can calculate the expected results on the client. I think calculating on the client is more efficient than hardcoding 200+ tests for every timezone.

kirmanak:
@rodionos  I thought the purpose of the tests is not to check the queries in every possible timezone, but to check that `WITH TIMEZONE` actually changes the query result to the expected one. I suppose, timezones are handled in one way, so there is no need to test all of them, two is enough, I guess. 
I check that the query results in `Asia/Kathmandu` and `Pacific/Marquesas` are different, so I can assume, that they differ in `Europe/Moscow` and `UTC` as well.
When we calculate test result, we can calculate it in incorrect way.
Please, take a look at the tests I've written by now, may be you'll change your opinion. Since all of them use `WITH TIMEZONE` keyword, the tests don't depend on ATSD timezone.

rodionos:
We can have both types of tests. Add hardcoded tests for Europe/Moscow, UTC, Asia/Kathmandu, Pacific/Marquesas, and a calculated test based on the current timezone in ATSD. We sure have such tests available in this project which take ATSD timezone into account. I looked at your tests - they're difficult to verify due to the usage of placeholders and separation of meaning  and logic between different methods and comments. I think they're testing what string does the server return. There is no date logic in them. With just 3 samples the hardcoded tests will not be reliable, because if the defect in interval condition is small, say 1 hour due to summer time offten, than this defect might not be reflected in query results. For example, what if `WHERE datetime >= '2018' AND datetime < '2019'` is incorrect by a small amount of time? A correct client-side calculation, on the other hand, doesn't leave the server a chance to make such as an error.

kirmanak:
@unrealwork If 3 samples as said @rodionos is not reliable, is it a good idea to add a list of samples with offset, for example, 20 minutes with interval, for example, from February to June?

kirmanak:
@rodionos @unrealwork  @raipc Since the code is difficult to read, I'm going to enumerate the performed tests: the tests are performed in five timezones: `Europe/Moscow`, `Asia/Kathmandu`, `UTC`, `Pacific/Marquesas` and the ATSD timezone. The samples are in interval from `2017-01-01T00:00:00Z` to `2018-07-01T00:00:00Z` with offset `20 minutes`.
Executed in `SELECT` clause functions: `date_format()`, `date_parse()`, `minute()`, `hour()`, `day()`, `extract (month from time)`, `is_weekday()`, `date_format(date_parse())`.
`WHERE` conditions: 

1. `datetime BETWEEN date_parse('2017-12-31T00:00:00', 'yyyy-MM-ddTHH:mm:ss')
AND date_parse('2018-01-03T03:00:00', 'yyyy-MM-ddTHH:mm:ss')`
2. `datetime BETWEEN '2017-12-31' AND '2018-01-03'`

`GROUP BY PERIOD()` periods: `1 day`, `1 week`, `1 month`.

Should I test anything additionally?

unrealwork:
@rodionos @kirmanak  We already have test for most of the queries specified in documentation.
The tests already have well organized datasets. In my opinion we should use following strategy.

1. Find appropriate feature in documentation
2. Find test for the feature
3. Add testcase using new syntax using existing series data.

> We should cover each fucntion anyway, if we want to cover all features in one test class it will be cumbersome and brain-overload.

kirmanak:
@rodionos Should I do it as @unrealwork suggested? If I should do it this way, should I add same set of tests(but use the new syntax)/replace current tests(if they don't test timezone casting)/add just one test?
For example, [SqlPeriodDataFilteringTest](https://github.com/axibase/atsd-api-test/blob/master/src/test/java/com/axibase/tsd/api/method/sql/function/period/filtering/SqlPeriodDataFilteringTest.java).
I can either add only one test, which is a copy of an existing test, but uses the new syntax or I can copy all of the present tests and use the new syntax or I can modify the tests, using the new syntax.

rodionos:
My proposal was outlined in #511#issuecomment-413155314. @unrealwork could comment on his proposal. The end result should be test coverage of all functions that depend on server time zone, including a case where the results are calculated based on the server timezone retrieved from the server at test time.

kirmanak:
@raipc @unrealwork Can we [fix](https://github.com/axibase/atsd-api-test/blob/f2506cd68fb3156ac13992ba2d45241c68a90ac8/pom.xml#L199) the flaky tests issue using [rerunFailingTestsCount](http://maven.apache.org/surefire/maven-surefire-plugin/examples/rerun-failing-tests.html)?

raipc:
@kirmanak The documentation states that this parameter is supported in JUnit provider only. `atsd-api-test` project uses TestNG

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
```diff
@@ -0,0 +1,293 @@
...
+                    ), testCase("is_weekday(time, 'USA')", timeZone,
+                    SAMPLES.stream().map(Sample::getZonedDateTime)
+                            .map((time) -> time.withZoneSameInstant(timeZone))
+                            .map(SqlWithTimezoneTest::isWeekday)
+                            .map(String::valueOf)
+                    ), testCase(
+                    "date_format(date_parse('2018-02-02T15:30:00', 'yyyy-MM-ddTHH:mm:ss'), 'yyyy-MM-ddTHH:mm:ss')",
+                    timeZone, SAMPLES.stream().map((sample) -> "2018-02-02T15:30:00")
+                    ), testCase("current_day", timeZone,
+                    SAMPLES.stream().map((sample) -> ZonedDateTime.now(timeZone))
```
raipc: Flaky tests, should be handled separately

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
```diff
@@ -0,0 +1,293 @@
...
+
+        SeriesMethod.insertSeriesCheck(series);
+    }
+
+    @DataProvider
+    public static Object[][] provideSelectExpressions() {
+        Object[][] results = null;
+        for (final ZoneId timeZone : timeZones) {
+            results = ArrayUtils.addAll(results, toArray(testCase("date_format(time, 'yyyy-MM-ddTHH:mm:ss')", timeZone,
+                    SAMPLES.stream().map(Sample::getZonedDateTime)
```
raipc: move common stream processing operations to a method

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
kirmanak: @raipc Which tests are flaky? What do you mean by `handled separately`: another data provider, another class?

raipc: `current_day` and `next_week` depend on current time, so to make the tests stable the lag between ATSD and test data preparator must be close to zero. It cannot be guaranteed.

For stability I would suggest to run such tests in a separate method, and if the output is not equal to expected, repeat the test once with recalculated expected values.

kirmanak: @raipc What is the size of such lag? I believe, it is between 24 hours (we start test at 00:01) and 0 (we start test at 23:59:59). But how often do we start test at this time? 
I think, that is not necessary. After pushing this commit I had to restart Travis CI check three times in a row because of an `ProcessingException`. For example, look at [this](https://travis-ci.org/axibase/atsd-api-test/builds/417239387) cron check. So, in my opinion, that is a common problem, which must be solved globally, not in this particular test.

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
kirmanak: @raipc moved to [`getTimeStream()`](https://github.com/axibase/atsd-api-test/blob/ae84e55a57787f7a7c937174567c605857fc099c/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java#L58)

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
raipc: ProcessingException is an ATSD problem caused by Jetty upgrade, it's not caused by error in test logic. Using of non-predictable time, in opposite, is a test smell. The tests are executed in different time zones, and it's not the situation when we can control that, for example, the tests won't be executed since 23:50 to 00:10 UTC.

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @BeforeClass
+    public static void prepareData() throws Exception {
+        final Series series = new Series(ENTITY_NAME, METRIC_NAME)
+                .addSamples(SAMPLES);
+
+        SeriesMethod.insertSeriesCheck(series);
+    }
+
+    private static Stream<ZonedDateTime> getTimeStream(final ZoneId timeZone) {
```
raipc: call it `modifyDate`, provide a Function<ZonedDateTime, Object> and return Stream<String>

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
```diff
@@ -0,0 +1,287 @@
...
+        final LocalDateTime monthEnd = LocalDateTime.parse("2018-02-01T00:00:00");
+        final LocalDateTime yearStart = LocalDateTime.parse("2017-01-01T00:00:00");
+        final LocalDateTime yearEnd = LocalDateTime.parse("2018-01-01T00:00:00");
+        Object[][] results = null;
+        for (final ZoneId timeZone : timeZones) {
+            results = ArrayUtils.addAll(results, toArray(
+                    testCase("datetime BETWEEN date_parse('2017-12-31T03:45:21', 'yyyy-MM-ddTHH:mm:ss') " +
+                                    "AND date_parse('2018-01-03T14:35:42', 'yyyy-MM-ddTHH:mm:ss')", timeZone,
+                            getTimeStream(timeZone)
+                                    .filter((time) -> time.compareTo(detailedStart.atZone(timeZone)) >= 0)
```
raipc: move to a function `between(from, to)`

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
```diff
@@ -0,0 +1,287 @@
...
+                                    .with(SECOND_OF_DAY, 0), summingInt((sample) -> sample.getValue().intValue())
+                            )).entrySet().stream().sorted(comparing(Map.Entry::getKey))
+                                    .map(Map.Entry::getValue).map(String::valueOf)
+                    )
+            ));
+        }
+        return results;
+    }
+
+    private static String listToString(final List<String> list) {
```
raipc: Use `StringUtils.join` or `String.join`

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
kirmanak: @raipc [That function](https://github.com/axibase/atsd-api-test/blob/0e54e79d34ef1da7a7d5fcfca75dc14223f08a2c/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java#L66) should accept a stream as the third arg, right?

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
kirmanak: @raipc Did you mean something like [it](https://github.com/axibase/atsd-api-test/blob/0e54e79d34ef1da7a7d5fcfca75dc14223f08a2c/src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java#L250)? That looks ugly, but I failed to invent a smarter way.

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
```diff
@@ -0,0 +1,305 @@
...
+
+    private static Stream<ZonedDateTime> between(final Stream<ZonedDateTime> stream,
+                                                 final ZonedDateTime from, final ZonedDateTime to) {
+        return stream.filter((time) -> time.compareTo(from) >= 0)
+                .filter((time) -> time.compareTo(to) <= 0);
+
+    }
+
+    private static <T> Stream<T> modifyDate(final Function<ZonedDateTime, T> function) {
+        return SAMPLES.stream().map(Sample::getZonedDateTime).map(function);
```
raipc: `time.withZoneSameInstant(timeZone)` and `String.valueOf` may be called in this method to lower the noise in data provider

kirmanak: @raipc Since `String.valueOf()` is not called as often as `modifyDate()` I did not move it, but I've moved the `withZoneSameInstant()` calls

kirmanak: @raipc Done. But I had to refuse from using the method in `provideWhereExpressions()`

src/test/java/com/axibase/tsd/api/method/sql/clause/with/SqlWithTimezoneTest.java
```diff
@@ -0,0 +1,307 @@
...
+    public void testCalendarKeywords(final String expression, final String timeZone, final String[][] expectedRows) {
+        final String query = String.format("SELECT %s FROM \"%s\" WITH TIMEZONE = \'%s\'",
+                expression, METRIC_NAME, timeZone
+        );
+        final String assertMessage =
+                String.format("Failed to modify timezone via WITH TIMEZONE in SELECT \"%s\"", expression);
+        try {
+            assertSqlQueryRows(assertMessage, expectedRows, query);
+        } catch (final AssertionError err) {
+            if (flakyRepeatCounter == 0) {
```
raipc: This approach will not work if more than one test fails because of transition to next day. Let's revert this change for now, I will implement retry later

PR #513, opened 2018-08-20
5600: Add escape characters in string functions tests

kirmanak:


src/test/java/com/axibase/tsd/api/method/sql/function/string/ReplaceTest.java
```diff
@@ -82,4 +82,16 @@ public void testReplaceWithDate() {
 
         assertSqlQueryRows(expectedRows, sqlQuery);
     }
+
+    @Issue("5600")
+    @Test
+    public void testReplaceLineBreak() {
+        String sqlQuery = "SELECT REPLACE('hello\nworld', '\n', ' ')";
```
rodionos: I'd like value that is being replaced to be retrieved from the database, and not passed as literal. Also, I'm curious how `\\n` will be handled. Add it to test as well.

rodionos:
Please search our docs on how we handle `\\` in string literals. Does it serve as an escaping character? Also search Oracle docs. How do they handle replacements of line feeds and is backslash used for escaping in their literal varchar data type?

kirmanak:
According to [our docs](https://axibase.com/docs/atsd/sql/#like-expression) backslash serves as the default escape character, but it is not specified if it is exclusively for wildcards or not.
According to [Oracle docs](https://docs.oracle.com/cd/B10501_01/text.920/a96518/cqspcl.htm#20741) backslash is used for escaping. 
To insert a carriage return/line feed they [offer](https://asktom.oracle.com/pls/asktom/f?p=100:11:0::::P11_QUESTION_ID:72412348057) use `chr()` function, which takes an ASCII code and returns the corresponding character.

rodionos:
Our docs are lacking clarity. Reach out to Alex T for information on how backslash is handled in string literal, other than in the LIKE expression where it escapes SQL wildcards.

rodionos:
Looks like we need to change the scope of this test. Add backslash tests of not just the REPLACE function, but all [string functions](https://axibase.com/docs/atsd/sql/#string-functions) as well as LIKE and REGEX operators for correct handling of the backslash.

* `\n`
* `\r`
* `\t`
* `\\`
* `\\n`
* `\\n\n`
* `\b`
* `\a`
* `\"`
* `\'`

For testing functions, insert data into the database containing these characters into fields: series tag value, entity tag value, series text annotation. Other fields may non support these non-printable characters.

Of relevance: https://axibase.com/docs/atsd/sql/#literals

src/test/java/com/axibase/tsd/api/method/sql/function/string/EscapeTest.java
```diff
@@ -0,0 +1,211 @@
...
+                        .toArray(String[][]::new)
+                ),
+                toArray("LENGTH(text)", Arrays.stream(CHARACTERS)
+                        .map((s) -> String.format(FORMAT, s).length())
+                        .map(String::valueOf)
+                        .map(ArrayUtils::toArray)
+                        .toArray(String[][]::new)
+                ),
+                toArray("CONCAT(text, 'Y')", Arrays.stream(CHARACTERS)
+                        .map((s) -> String.format(FORMAT, s).concat("Y"))
```
unrealwork: You shouldn't write another code doing the same stuff as the ATSD does. Just provide constants, instead of calculating the expected result.

kirmanak: @unrealwork fixed. Please, review again

src/test/java/com/axibase/tsd/api/method/sql/function/string/EscapeTest.java
```diff
@@ -0,0 +1,337 @@
...
+                .addSamples(Sample.ofDateText("2018-08-20T03:00:00.000Z", "hello\\world"))
+                .addSamples(Sample.ofDateText("2018-08-20T04:00:00.000Z", "hello\\nworld"))
+                .addSamples(Sample.ofDateText("2018-08-20T05:00:00.000Z", "hello\\n\nworld"))
+                .addSamples(Sample.ofDateText("2018-08-20T06:00:00.000Z", "hello\bworld"))
+                .addSamples(Sample.ofDateText("2018-08-20T07:00:00.000Z", "hello\"world"))
+                .addSamples(Sample.ofDateText("2018-08-20T08:00:00.000Z", "hello\'world"))
+                .addSamples(Sample.ofDateText(
+                        "2018-08-20T09:00:00.000Z", String.format("hello%sworld", ALARM_CHARACTER))
+                )
+                .addTag(String.valueOf("\n".hashCode()), "hello\nworld")
```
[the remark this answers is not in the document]
kirmanak: @unrealwork done

PR #523, opened 2018-11-16
Feature #5770. SQL: pi function

Sadredneck:


src/test/java/com/axibase/tsd/api/method/sql/function/math/SqlPiFunctionTest.java
```diff
@@ -0,0 +1,110 @@
...
+
+        StringTable resultTable = queryResponse(sqlQuery).readEntity(StringTable.class);
+
+        Assert.assertEquals(resultTable.getRows().get(0).get(0), expectedValue, "Radians value differed");
+    }
+
+    @Issue("5770")
+    @Test(dataProvider = "provideFunctionNames")
+    public void testNullValues(String functionName) {
+        String sqlQuery = String.format("SELECT %s(null)",
```
raipc: SELECT SIN(CAST(null AS NUMBER))

Sadredneck: Fixed

src/test/java/com/axibase/tsd/api/method/sql/function/math/SqlPiFunctionTest.java
```diff
@@ -0,0 +1,105 @@
...
+
+    @DataProvider
+    public static Object[][] provideDegreesValues() {
+        return new Object[][]{{"pi()/4", "45.0"}, {"pi()/2", "90.0"}, {"pi()*3/4", "135.0"}, {"pi()", "180.0"},
+                {"pi()*5/4", "225.0"}, {"pi()*3/2", "270.0"}, {"pi()*7/4", "315.0"}, {"pi()*2", "360.0"}};
+    }
+
+    @DataProvider
+    public static Object[][] provideRadianValues() {
+        return new Object[][]{{"45.0", "0.7853981633974483"}, {"90.0", "1.5707963267948966"},
```
raipc: magic numbers

Sadredneck: Changed to Math functions

rodionos:
I'd rename this class to SqlRadianFunctionTest. Do you test for the correct behavior if the argument is not valid?

rodionos:
* One class for both radians and degrees is sufficient.
* By invalid arguments I mean real numbers that are outside of "область определения функции". How do the functions behave , by the way, when the argument is not valid. Do they raise exception or do they return `Double.NaN`?

Sadredneck:
I don't think there's a number we can't apply to degrees or radians functions. Invalid arguments return NaN.

rodionos:
Ok, while radians and degrees functions accept any real number, this may not be the case in general, e.g. asin(2). Something to cover in the remaining trig function tests.

PR #531, opened 2019-04-25
Added new test for different options of horizon.

VVVyacheslav:


src/test/java/com/axibase/tsd/api/method/series/SeriesQueryHorizonTest.java
```diff
@@ -38,25 +40,16 @@
 public class SeriesQueryHorizonTest extends SeriesMethod{
 
     @Data
-    @NoArgsConstructor
+    @RequiredArgsConstructor
     @Accessors(chain = true)
```
VeselovAlex: This annotation seems to be redundant after transition to `final` fields

PR #535, opened 2019-05-06
Jersey upgrade

raipc:
Reasons to this refactoring: multiple warnings from snyk.io, uncomfortable way to access config singleton

### DIFF
--- a/src/main/java/com/axibase/tsd/api/method/MethodParameters.java
+++ b/src/main/java/com/axibase/tsd/api/method/MethodParameters.java
@@ -3,10 +3,8 @@ package com.axibase.tsd.api.method;
 import com.fasterxml.jackson.databind.ObjectMapper;
 
 import javax.ws.rs.client.WebTarget;
-import java.util.Collection;
-import java.util.Map;
-import java.util.Stack;
-import java.util.StringJoiner;
+import java.util.*;
+import java.util.stream.Collectors;
 
 public abstract class MethodParameters {
     private static final ObjectMapper MAPPER = new ObjectMapper();
@@ -17,7 +15,7 @@ public abstract class MethodParameters {
     }
 
     @SuppressWarnings("unchecked")
-    private static WebTarget appendMap(WebTarget target, Stack<String> path, Map<String, Object> map) {
+    private static WebTarget appendMap(WebTarget target, Deque<String> path, Map<String, Object> map) {
         for (Map.Entry<String, Object> entry : map.entrySet()) {
             Object value = entry.getValue();
             if (value != null) {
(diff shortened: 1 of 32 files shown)


src/main/java/com/axibase/tsd/api/method/MethodParameters.java
```diff
@@ -43,19 +41,15 @@ private static String formatParameterName(Iterable<String> parameterName) {
         return String.join(".", parameterName);
     }
 
+    @SuppressWarnings("unchecked")
     private static String formatCollection(Collection collection) {
```
unrealwork: Maybe the variant is more readable

```java
private static String formatCollection(Collection<?> collection) {
        return collection.stream()
                .filter(Objects::nonNull)
                .map(Object::toString)
                .collect(Collectors.joining(","));

    }
```

src/main/java/com/axibase/tsd/api/method/csv/CSVUploadMethod.java
```diff
@@ -26,38 +27,45 @@ public static File resolvePath(String path) throws URISyntaxException, FileNotFo
...
-        MultiPart multiPart = new MultiPart();
-        MediaType mediaType = getMediaTypeFromFile(file);
-        if (mediaType.equals(MediaType.APPLICATION_OCTET_STREAM_TYPE)) {
-            mediaType = new MediaType("text", "csv");
-        }
-        FileDataBodyPart fileDataBodyPart = new FileDataBodyPart("filedata", file, mediaType);
-        multiPart.bodyPart(fileDataBodyPart);
+        try (MultiPart multiPart = new MultiPart()) {
+            MediaType mediaType = getMediaTypeFromFile(file);
+            if (mediaType.equals(MediaType.APPLICATION_OCTET_STREAM_TYPE)) {
```
unrealwork: Flip the `equals` for null-safety.

src/main/java/com/axibase/tsd/api/transport/tcp/TCPSender.java
```diff
@@ -20,41 +20,27 @@
...
              DataOutputStream requestStream = new DataOutputStream(socket.getOutputStream());
              BufferedReader responseStream = new BufferedReader(
                      new InputStreamReader(socket.getInputStream(), StandardCharsets.UTF_8)
              )) {
-            String request = isDebugMode ? DEBUG_PREFIX.concat(command) : command;
+
             IOUtils.write(request, requestStream);
             String response = isDebugMode ? responseStream.readLine() : null;
-            LOGGER.debug(" > tcp://{}:{} \n\t{}\n< {}", config.getServerName(), config.getTcpPort(), request, response);
+            LOGGER.debug(" > tcp://{}:{} \n\t{}\n< {}", host, port, request, response);
```
unrealwork: We should choose the name convention for loggers. Better migrate to Lombok as in other classes.

src/test/java/com/axibase/tsd/api/util/TestUtil.java
```diff
@@ -248,6 +248,10 @@ public static String quoteEscape(String s) {
         return Arrays.stream(data).map(o -> new Object[] {o}).toArray(Object[][]::new);
     }
 
+    public static Object[][] convertTo2DimArray(Collection<?> data) {
+        return data.stream().map(o -> new Object[] {o}).toArray(Object[][]::new);
```
unrealwork: You can use `ArrayUtils::toArray` from Apache Commons instead raw array creation

src/main/java/com/axibase/tsd/api/transport/tcp/TCPSender.java
raipc: I didn't modify the logger name. It should be done in the other PR

Issue #537, opened 2019-05-30
Add styleguide


raipc:
We definitely need a styleguide to simplify code review process. Currently we have many collisions when different developers tend to put their own code practice into the reviewed code. Such collisions must be resolved centrally, by creating rules in the style guide.
I would suggest to try to adopt some existing style guide, e.g. Google's one, and then add/remove our own rules after team discussion.

annstriganova:
Idea supports style guide exporting/importing, when we come to an agreement, we can create `.xml` and add it to repo.
![Screenshot_2019-05-31_16-44-30](https://user-images.githubusercontent.com/20650874/58709950-e6780f00-83c3-11e9-8abc-21c102688a52.png)

PR #538, opened 2019-05-30
Style guide

raipc:
#537 Please, review the style guide and suggest rules/improvements

### DIFF
--- a/STYLEGUIDE.md
+++ b/STYLEGUIDE.md
@@ -1,24 +1,103 @@
 # Axibase Style Guide for Java Code
 
-## 1 Introduction
-
 This document serves as the definition of Axibase coding standards for source code in the Java™ Programming Language. It's mostly based on [Google Code Style](https://google.github.io/styleguide/javaguide.html) with some stricter rules covering tests implementation.
 
-### 1.1 Terminology notes
+---
+
+## Contents
+
+* [1 Terminology notes](#1-terminology-notes)
+* [2 Source file basics](#2-source-file-basics)
+  * [2.1 File name](#21-file-name)
+  * [2.2 File encoding: UTF-8](#22-file-encoding-utf-8)
+  * [2.3 Special characters](#23-special-characters)
+    * [2.3.1 Whitespace characters](#231-whitespace-characters)
+    * [2.3.2 Special escape sequences](#232-special-escape-sequences)
+    * [2.3.3 Non-ASCII characters](#233-non-ascii-characters)
+* [3 Source file structure](#3-source-file-structure)
+  * [3.1 License or copyright information, if present](#31-license-or-copyright-information-if-present)
+  * [3.2 Package statement](#32-package-statement)
+  * [3.3 Import statements](#33-import-statements)
+    * [3.3.1 Wildcard imports](#331-wildcard-imports)
+    * [3.3.2 No line-wrapping](#332-no-line-wrapping)
+    * [3.3.3 Ordering and spacing](#333-ordering-and-spacing)
+    * [3.3.4 No static import for classes](#334-no-static-import-for-classes)
+    * [3.3.5 Prefer static imports for Assert.* methods, non-static imports otherwise](#335-prefer-static-imports-for-assert-methods-non-static-imports-otherwise)
+    * [3.3.6 Unused imports must be deleted](#336-unused-imports-must-be-deleted)
+  * [3.4 Class declaration](#34-class-declaration)
+    * [3.4.1 Exactly one top-level class declaration](#341-exactly-one-top-level-class-declaration)
+    * [3.4.2 Ordering of class contents](#342-ordering-of-class-contents)
+      * [3.4.2.1 Overloads: never split](#3421-overloads-never-split)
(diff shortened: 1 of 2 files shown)


raipc:
Suggestion.
3.4.2.2. Private methods placement
Private methods must be placed under the first method they are called from.

raipc:
Allow wildcard imports

raipc:
Suggestion.
All new model classes must be immutable. Reason of introduced immutability must be properly documented.

STYLEGUIDE.md
```diff
@@ -0,0 +1,726 @@
...
+
+### 3.2 Package statement
+
+The package statement is **not line-wrapped**. The column limit (Section 4.4, [Column limit: 100](#s4.4-column-limit)) does not apply to package statements.
+
+### 3.3 Import statements
+
+#### 3.3.1 No wildcard imports
+
+**Wildcard imports**, static or otherwise, **are not used**.
```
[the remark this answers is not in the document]
raipc: I think it's not important whether import is wildcard or not. It is always generated by the IDE.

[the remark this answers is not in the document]
unrealwork: I'm sure that a big amount of imports is bad practice. Most likely the class is overcomplicated.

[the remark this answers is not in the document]
raipc: Good luck to refactor all of it

STYLEGUIDE.md
```diff
@@ -0,0 +1,769 @@
...
+What is important is that each class uses **_some_ logical order**, which its maintainer could explain if asked. For example, new methods are not just habitually added to the end of the class, as that would yield "chronological by date added" ordering, which is not a logical ordering.
+
+##### 3.4.2.1 Overloads: never split
+
+When a class has multiple constructors, or multiple methods with the same name, these appear sequentially, with no other code in between (not even private members).
+
+##### 3.4.2.2. Private methods placement
+
+The private method must be placed under the first method it is called from. Multiple private methods used in one place must be placed in order of usage.
+
```
VVVyacheslav: But it is true for tests too?

raipc: I added this rule specifically for tests

VVVyacheslav: But then get the following situation: test, next the private method, then other tests.

VeselovAlex: IMHO we should place the most important content – tests and test data, first. It is the actual reason to view class source code. If you place helpers after the place of their first usage, you will probably have to scroll both sides^ up and down^ to navigate between helper body and the caller method. 

The common practice is to place public members to the top since they are the members of class interface, not implementation details.

STYLEGUIDE.md
```diff
@@ -0,0 +1,769 @@
...
+    *   the ampersand in a conjunctive type bound: `<T extends Foo & Bar>`
+    *   the pipe for a catch block that handles multiple exceptions: `catch (FooException | BarException e)`
+    *   the colon (`:`) in an enhanced `for` ("foreach") statement
+    *   the arrow in a lambda expression: `(String str) -> str.length()`
+    
+    but not
+    *   the two colons (`::`) of a method reference, which is written like `Object::toString`
+    *   the dot separator (`.`), which is written like `object.toString()`
+5.  After `,:;` or the closing parenthesis (`)`) of a cast
+6.  On both sides of the double slash (`//`) that begins an end-of-line comment. Here, multiple spaces are allowed, but not required.
```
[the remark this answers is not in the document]
VeselovAlex: This rule is apllied for comments in the end of code line:

Correct:
```java
long nextId = getNextId(); // Request next id from key manager
```

Bad:
```java
long nextId = getNextId();//Request next id from key manager
```

STYLEGUIDE.md
raipc: If you place private methods generating test data to the bottom you don't know what is the data and need to scroll down.
> The common practice is to place public members to the top since they are the members of class interface, not implementation details.
Never heard of this practice, and test classes don't implement interfaces

raipc: I think best way of methods ordering is one where you can read entire code from top to bottom in Github UI, without IDE

KatrineTY: It is reasonable if there are no public methods that use the same private methods. How should we place methods in that case?

VeselovAlex: Some examples:

raipc: Oracle
```
These methods should be grouped by functionality rather than by scope or accessibility. 
For example, a private class method can be in between two public instance methods. 
The goal is to make reading and understanding the code easier.
```

STYLEGUIDE.md
```diff
@@ -0,0 +1,996 @@
...
+#### 3.3.2 No line-wrapping
+
+Import statements are **not line-wrapped**. The [column limit](#44-column-limit) does not apply to import statements.
+
+#### 3.3.3 Ordering and spacing
+
+Imports are ordered as follows:
+
+1. All static imports in a single block.
+2. All non-static imports in a single block.
```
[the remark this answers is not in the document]
raipc: Yes, you are right

Issue #542, opened 2019-06-10
Update Mocks name generation


Arthasasyan:
Current name generator create similar names for every test iteration. It can be a problem if you need to run tests several times as you need to delete what was generated every time. I suggest the following changes to `TestNameGenerator` class to guarantee unique values:

```java
String newTestName(Key key) {
        String namePrefix = getPrefix(key);
        int testNumber = prefixDictionary.computeIfAbsent(namePrefix, prefix -> new AtomicInteger((int)System.currentTimeMillis()))
                .incrementAndGet();
        return namePrefix + "-" + testNumber;
    }
```

Arthasasyan:
Probably, `new Random().nextInt()` could be a better decision to avoid type conversions from long to int

raipc:
Why not just add a suffix with a time of test config initialization?

Arthasasyan:
We can add new field to TestNameGenerator class

```java
private static final String TEST_INITIALIZATION_TIME = Util.ISOFormat(System.currentTimeMillis());
```

And then rewrite function like this 

```java
String newTestName(Key key) {
        String namePrefix = getPrefix(key);
        int testNumber = prefixDictionary.computeIfAbsent(namePrefix, prefix -> new AtomicInteger(0))
                .incrementAndGet();
        return namePrefix + "-" + testNumber + "-" + TEST_INITIALIZATION_TIME;
    }
```

PR #543, opened 2019-06-11
Mocks name update

Arthasasyan:
Changed naming strategy due to issue #542 on github


Arthasasyan:
Before changing strategy, `testMetaNonEmptyTagsExpansion` could throw an error because metrics were in different order

src/test/java/com/axibase/tsd/api/method/sql/clause/join/JoinWithAtsdSeriesTest.java
```diff
@@ -336,7 +336,7 @@ SELECT isnull(t1.datetime, t2.datetime) as \"date\", t1.value, t2.value
     }
 
     @Issue("4089")
-    @Test
+    @Test(enabled = false)
```
raipc: Do not disable passing tests

Arthasasyan: These tests were disabled in master. I probably had a checkout error end `enabled=false` flags were not presented in a new branch 
https://github.com/axibase/atsd-api-test/blob/7ba9645ae26bb3192e5d200e940e2ad6466814c1/src/test/java/com/axibase/tsd/api/method/sql/clause/join/JoinWithAtsdSeriesTest.java#L339
https://github.com/axibase/atsd-api-test/blob/7ba9645ae26bb3192e5d200e940e2ad6466814c1/src/test/java/com/axibase/tsd/api/method/sql/syntax/SqlSubqueryTest.java#L97

src/test/java/com/axibase/tsd/api/method/sql/meta/MetaDataTest.java
```diff
@@ -320,8 +320,26 @@ public void testMetaNonEmptyTagsExpansion() throws Exception {
                 "string"
         };
 
-        assertSqlMetaNamesAndTypes("Wrong /api/sql/meta tags expansion for multiple existent metrics from atsd_series",
-                expectedNames, expectedTypes, sqlQuery);
+        try {
```
raipc: The order must not be changed after suffix adding. In case of flaky assertion error the query must be fixed in another pull request

Arthasasyan:
Pending for #549 to be resolved

PR #550, opened 2019-06-20
Debug mode test

Arthasasyan:
Test how network commands work with debug prefix due to issue 6319


src/test/java/com/axibase/tsd/api/method/metric/transport/tcp/MetricTCPTest.java
```diff
@@ -0,0 +1,203 @@
...
+                metric.getVersioned()
+        );
+
+        assertMetricExisting(assertMessage, metric);
+    }
+
+
+    @Issue("6319")
+    @Test
+    public void testTimezone() throws Exception {
```
raipc: What does this test do?

src/test/java/com/axibase/tsd/api/method/metric/transport/tcp/MetricTCPTest.java
```diff
@@ -0,0 +1,203 @@
...
+                metric.getInterpolate()
+        );
+
+        assertMetricExisting(assertMessage, metric);
+    }
+
+    @DataProvider(name = "incorrectTimeZoneProvider")
+    public Object[][] provideIncorrectTimeZoneData() {
+        return new Object[][]{
+                {"a"},
```
raipc: The values must be one of following groups:
* unknown timezone
* unsupported timezone (there are some 3-character abbreviations)
* illegal GMT offset

src/test/java/com/axibase/tsd/api/method/message/transport/tcp/MessageTCPTest.java
```diff
@@ -0,0 +1,52 @@
...
+        message.setDate(TestUtil.getCurrentDate());
+
+        PlainCommand command = new MessageCommand(message);
+        assertGoodTcpResponse("Message was not sent via TCP", TCPSender.send(command, true));
+        assertMessageExisting("Message was not sent via TCP", message);
+    }
+
+    @Issue("6319")
+    @Test
+    public void testMalformedRequest() throws Exception {
```
raipc: Mention why it is malformed, in name or description

src/main/java/com/axibase/tsd/api/transport/tcp/TCPSender.java
```diff
@@ -70,4 +71,28 @@ private static String buildRequest(Collection<? extends PlainCommand> commands)
     public static void send(Collection<? extends PlainCommand> commands) throws IOException {
         send(commands, false);
     }
+
+    public static void assertBadTcpResponse(String response) {
```
unrealwork: Are the assertion methods related to the class? Please move it corresponding Assertion Class

src/test/java/com/axibase/tsd/api/method/message/transport/tcp/MessageTCPTest.java
```diff
@@ -0,0 +1,52 @@
...
+        Message message = new Message(Mocks.entity().replaceAll("-", " "), "logger");
+        message.setMessage(Mocks.message());
+        message.setDate(TestUtil.getCurrentDate());
```
unrealwork: Use static date for the test. If it will fail, It can be not reproducible.

src/test/java/com/axibase/tsd/api/method/metric/transport/tcp/MetricTCPTest.java
Arthasasyan: This test had to check that metric with correct timezone is inserted. An error with filter expression is also made at MetricCommand test. Both will be fixed

src/test/java/com/axibase/tsd/api/method/metric/transport/tcp/MetricTCPTest.java
Arthasasyan: It seems that ATSD does not support setting timezone as GMT offset
I changed this dataprovider for tcp and command tests

src/test/java/com/axibase/tsd/api/method/message/transport/tcp/MessageTCPTest.java
Arthasasyan: Done

src/main/java/com/axibase/tsd/api/transport/tcp/TCPSender.java
Arthasasyan: Should it be moved to `TCPSenderTest` class?

src/test/java/com/axibase/tsd/api/method/message/transport/tcp/MessageTCPTest.java
Arthasasyan: Done

src/test/java/com/axibase/tsd/api/method/entity/EntityCommandTest.java
```diff
@@ -61,7 +60,7 @@ public void testUpdateEntityTagsForExistEntity() throws Exception {
     @Issue("3111")
 
     @Test
-    public void testAddNewEntityTagsMailformedForNewEntity() throws Exception {
+    public void testAddNewEntityTagsMalformedForNewEntity() throws Exception {
```
raipc: I think there is too much duplication: tests for TCP and HTTP transport differ in only 1 line. It's too difficult to support them in sync. Create an enum `Transport {HTTP, TCP}` and parameterize the test class instance using factory. Add transport to assertion message to differentiate cases.

```java
@RequiredArgsConstructor
public class EntityCommandTest extends EntityTest {
    ...
    private final Transport transport;

    public static class TestFactory {
        @Factory
        public Object[] factoryMethod() {
            return new Object[]{new EntityCommandTest(Transport.HTTP), new EntityCommandTest(Transport.TCP)};
        }
    }
    ....
```

unrealwork: @Arthasasyan @raipc You can also annotate the constructor with `@Factory` and provide `@DataProvider`. For me, it's less complicated.

```java
@Factory(dataProvider = "transport")
    public EntityCommandTest(final Transport transport) {
        this.transport = transport;
    }

    @DataProvider
    private static Object[][] transport() {
        return convertTo2DimArray(Transport.values());
    }
```

Arthasasyan:
Should `StringCommand` be renamed to `SimpleCommand` as it is in [atsd-api-java](https://github.com/axibase/atsd-api-java/blob/master/src/main/java/com/axibase/tsd/network/SimpleCommand.java)

src/test/java/com/axibase/tsd/api/method/entity/EntityCommandTest.java
```diff
@@ -28,56 +34,85 @@
...
         storedEntityWithTags.addTag(E_TAG_1, E_VAL_1);
         createOrReplaceEntityCheck(storedEntityWithTags);
         storedEntityWithTags.addTag(E_TAG_2, E_VAL_2);
         PlainCommand command = new EntityCommand(storedEntityWithTags);
-        CommandMethod.send(command);
-        assertEntityExisting("Entity tag isn't add for existing entity",
-                storedEntityWithTags
-        );
+        if(transport == Transport.HTTP) {
+            CommandMethod.send(command);
```
raipc: Move to Transport enum

Arthasasyan: @raipc do you mean adding methods like these?
```java
public void assertGood(PlainCommand command) {
if(this == Transport.HTTP) {
CommandMethod.send(command);
} else if (this == Transport.TCP) {
        );
            assertGoodTcpResponse(TCPSender.send(command, true));
        }
}
```

```java
public void assertBad(PlainCommand command) {
if(this == Transport.HTTP) {
        assertEquals(expectedResult, CommandMethod.send(command));
            CommandSendingResult expectedResult = new CommandSendingResult(1, 0);
            assertEquals(expectedResult, CommandMethod.send(command));
        } else if(this == Transport.TCP) {
            assertBadTcpResponse(TCPSender.send(command, true));
        }
}
```

raipc:
```java
public enum Transport {
    HTTP {
        @Override
        public void sendAndCompareToExpected(PlainCommand command, CommandSendingResult expected, String message) {
            assertEquals(message, expected, CommandMethod.send(command));
        }
    },
    TCP {
        @Override
        public void sendAndCompareToExpected(PlainCommand command, CommandSendingResult expected, String message) throws IOException {
            final String response = TCPSender.send(command, true);
            if (expected.getFail() > 0) {
                assertBadTcpResponse(response, message);
            } else {
                assertGoodTcpResponse(response, message);
            }
        }
    };

    public abstract void sendAndCompareToExpected(PlainCommand command, CommandSendingResult expected, String message) throws IOException;
}
```

src/test/java/com/axibase/tsd/api/method/entity/EntityCommandTest.java
```diff
@@ -27,118 +31,141 @@
...
+        Entity storedEntityWithTags = new Entity(Mocks.entity());
         storedEntityWithTags.addTag(E_TAG_1, E_VAL_1);
         createOrReplaceEntityCheck(storedEntityWithTags);
         storedEntityWithTags.addTag(E_TAG_2, E_VAL_2);
         PlainCommand command = new EntityCommand(storedEntityWithTags);
-        CommandMethod.send(command);
-        assertEntityExisting("Entity tag isn't add for existing entity",
-                storedEntityWithTags
-        );
+        String assertMessage = "Entity tag isn't added for existing entity with " + transport.toString();
```
raipc: Don't call toString explicitly

src/test/java/com/axibase/tsd/api/transport/Transport.java
```diff
@@ -0,0 +1,34 @@
...
+import java.io.IOException;
+
+import static com.axibase.tsd.api.transport.tcp.TCPSenderTest.assertBadTcpResponse;
+import static com.axibase.tsd.api.transport.tcp.TCPSenderTest.assertGoodTcpResponse;
+import static org.testng.AssertJUnit.assertEquals;
+
+public enum Transport {
+    HTTP {
+        @Override
+        public void sendAndCompareToExpected(PlainCommand command, CommandSendingResult expected, String message) {
```
unrealwork: @raipc @Arthasasyan I'm not sure that assertions should be in `Transport`. Transport should have only responsibilities related to sending a network command.


```java
public enum Transport {
    HTTP {
        @Override
        public boolean send(final PlainCommand command) throws IOException {
            final CommandSendingResult result = CommandMethod.send(command);
            return result.getFail() == 0;
        }
    },
    TCP {
        @Override
        public boolean send(final PlainCommand command) throws IOException {
            final String response = TCPSender.send(command, true);
            return "ok".equals(response);
        }
    };

    public abstract boolean send(PlainCommand command) throws IOException;
}
```

And then  sending assertion will be simplified

```java
// assertCommandIsSent
assertTrue(assertMessage, transport.send(command));

// assertCommandIsNotSent
assertFalse(assertMessage, transport.send(command));
```

src/test/java/com/axibase/tsd/api/method/entity/EntityCommandTest.java
```diff
@@ -27,118 +29,141 @@
...
 
-    @Issue("3111")
+    private final Transport transport;
+
+    @Factory(dataProvider = "transport")
+    public EntityCommandTest(final Transport transport) {
+        this.transport = transport;
+    }
 
+    @DataProvider
```
unrealwork: Duplicated `@DataProvider` move it to the `Transport` or more appropriate class and reuse it via `dataProviderClass` param of `@DataProvider` annotation.

```java
@Factory(dataProvider = "transport", dataProviderClass = Transport.class)
    public EntityCommandTest(final Transport transport) {
        this.transport = transport;
    }
```

src/test/java/com/axibase/tsd/api/transport/Transport.java
Arthasasyan: I believe that these methods should have another name because they are `boolean`. I suggest `isSent`

unrealwork: Commonly `is` prefix in Java is used for getters that return boolean type, the methods shouldn't contains actions only result of action. This case likes the creation of files in old Java API - [createNewFile](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#createNewFile())

Arthasasyan: Renamed with the latest commit

src/test/java/com/axibase/tsd/api/transport/tcp/TCPSenderTest.java
```diff
@@ -33,4 +34,28 @@ public void testDebugSend() throws Exception {
         assertEquals("ok", result);
         assertCheck(new EntityCheck(entity));
     }
+
+    public static void assertBadTcpResponse(String response) {
```
unrealwork: Other usages of these assert messages could be replaced via assertion of a result of `TCP.send(command)` method.

Arthasasyan: Currently most of old tests divide similar command tests into different classes. For example, [BackslashCharEscapeTest](https://github.com/axibase/atsd-api-test/blob/master/src/test/java/com/axibase/tsd/api/method/series/command/BackslashCharEscapeTest.java) and [DQuoteCharEscapeTest](https://github.com/axibase/atsd-api-test/blob/master/src/test/java/com/axibase/tsd/api/method/series/command/DQuoteCharEscapeTest.java). When writing tests for TCP transport, I tried to escape it by adding all of special characters to one case like this
https://github.com/axibase/atsd-api-test/blob/2d5fd995b2af1af53690608686cc0294c76cab32/src/test/java/com/axibase/tsd/api/method/series/transport/tcp/SeriesTCPTest.java#L47
Could you specify what should be done?:
1) Written TCP tests should be deleted and existing command tests should be edited so that they now support both HTTP and TCP methods.
2) Written TCP test should be modified so that they will use `Transport.TCP.send(command)`method
Or something else?

unrealwork: I've written just about usages of methods `assertGoodTcpResponse`, `assertBadTcpResponse` which can be rewritten as:

```java
assertGoodTcpResponse(TCPSender.send(command, true));
// ->
assertTrue(TCP.send(command));
```

src/test/java/com/axibase/tsd/api/transport/tcp/TCPSenderTest.java
```diff
@@ -13,6 +13,7 @@
 import static com.axibase.tsd.api.util.Mocks.*;
 import static java.util.Collections.singletonList;
 import static org.testng.AssertJUnit.assertEquals;
+import static org.testng.AssertJUnit.fail;
```
unrealwork: Unused import

src/main/java/com/axibase/tsd/api/model/command/StringCommand.java
```diff
@@ -0,0 +1,18 @@
+package com.axibase.tsd.api.model.command;
+
+/**
+ * Class that transforms string command to PlainCommand
+ */
+public class StringCommand extends AbstractCommand {
+    private final String command;
+
+    public StringCommand(String command) {
+        super("");
```
unrealwork: Reuse existing `StringBuilder` from `AbstractCommand`

```java
public class StringCommand extends AbstractCommand {

    public StringCommand(final String commandText) {
        super(commandText);
    }

    @Override
    public String compose() {
        return commandBuilder().toString();
    }
}
```

src/test/java/com/axibase/tsd/api/method/series/transport/tcp/SeriesTCPTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static Date TEST_DATE = Util.parseDate("2019-06-20T12:00:00.000Z");
+
+    @Issue ("6319")
+    @Test
+    public void testNormalWorkflow() throws Exception {
+        Series series = new Series(Mocks.entity(), Mocks.metric());
+        series.addSamples(Sample.ofJavaDateInteger(TEST_DATE, 22));
+        PlainCommand command = new SeriesCommand(series);
+        assertTrue(Transport.TCP.send(command));
```
raipc: Please, parameterize transport here as well

src/main/java/com/axibase/tsd/api/model/command/SeriesCommand.java
```diff
@@ -33,68 +43,42 @@ public SeriesCommand(Map<String, String> texts, Map<String, String> values, Stri
...
-    public Map<String, String> getTexts() {
-        return texts;
-    }
-
-    public void setTexts(Map<String, String> texts) {
-        this.texts = texts;
-    }
+    public SeriesCommand(Series series) {
+        super(SERIES_COMMAND);
+        if(!series.getData().isEmpty()) {
```
raipc: Please, use Reformat Code feature. See [styleguide](https://github.com/axibase/atsd-api-test/blob/styleguide/STYLEGUIDE.md#462-horizontal-whitespace)

src/test/java/com/axibase/tsd/api/method/message/transport/tcp/MessageTCPTest.java
```diff
@@ -0,0 +1,58 @@
...
+
+    @Issue("6319")
+    @Test
+    public void testNormalWorkflow() throws Exception {
+        Message message = new Message(Mocks.entity(), "logger");
+        message.setMessage(Mocks.message());
+        message.setDate(TEST_DATE);
+
+        PlainCommand command = new MessageCommand(message);
+        assertTrue("Message was not sent via TCP", Transport.TCP.send(command));
```
raipc: Please, parameterize transport here as well

PR #560, opened 2019-07-12
Series query API - test 'Evaluation'.

MikhailZvagelsky:
The 'evaluation' series transformation is not documented yet, and the related ticket #6153 is too messy to link it. This doesn't test the 'evaluation' itself, but checks how it is built in other transformations sequence.


src/test/java/com/axibase/tsd/api/method/series/PostEvaluationTest.java
```diff
@@ -0,0 +1,348 @@
...
+                .addType(AggregationType.COUNT)
+                .addType(AggregationType.SUM)
+                .addType(AggregationType.MAX);
+        SeriesQuery query = baseQuery
+                .withAggregate(aggregationSettings)
+                .withTransformationOrder(Arrays.asList(EVALUATE, AGGREGATE));
+        List<Series> seriesList = querySeriesAsList(query);
+        assertEquals(seriesList.size(), 3 * seriesCount);
+        CommonAssertions.assertSeriesSize(seriesList, 2);
+        CommonAssertions.checkValues(seriesList,
```
unrealwork: Could be replaced with two-dimension array
```java
CommonAssertions.checkValues(seriesList, new String[][]{
                {"12", "8"},
                {"3", "2"},
                {"10", "40"},
                {"5", "20"},
                {"20", "45"},
                {"10", "22.5"}
        });
```

src/test/java/com/axibase/tsd/api/util/CommonAssertions.java
```diff
@@ -143,4 +150,124 @@ public static void assertDecimals(final BigDecimal expected, final BigDecimal ac
...
+     */
+    public static void checkValues(Series series, String... values) {
+        BigDecimal[] decimals = Arrays.stream(values).map(BigDecimal::new).toArray(BigDecimal[]::new);
+        if (!hasValues(series, decimals)) {
+            throw new AssertionError(String.format("Array %n%s%n doesn't equal to array of series values %n%s%n",
+                    Arrays.toString(values), series));
+        }
+    }
+
+    /** True iff specified array is array of the series values (in the same order). */
```
unrealwork: `iff` - misspelling

src/test/java/com/axibase/tsd/api/util/CommonAssertions.java
```diff
@@ -143,4 +150,124 @@ public static void assertDecimals(final BigDecimal expected, final BigDecimal ac
...
+    public static boolean hasValues(Series series, BigDecimal... valuesArray) {
+        if (series.getData().size() != valuesArray.length) {
+            return false;
+        }
+        int index = 0;
```
unrealwork: Indexed iteration could be implemented using `Stream`.
```java
return IntStream.range(0, valuesArray.length)
                .allMatch(i -> valuesArray[i].compareTo(data.get(i).getValue()) == 0);
```

src/test/java/com/axibase/tsd/api/util/CommonAssertions.java
```diff
@@ -143,4 +150,124 @@ public static void assertDecimals(final BigDecimal expected, final BigDecimal ac
...
+     * such that the array consists of all the series values.
+     * @param seriesList List of series.
+     * @param values     Collection of values arrays. Each value will be converted to {@link BigDecimal}.
+     *                   (Strings are used to simplify method usage.)
+     */
+    public static void checkValues(List<Series> seriesList, String[]... values) {
+        String msg = "Array of values %s not found in series list.";
+        for (String[] valuesArray : values) {
+            BigDecimal[] decimals = Arrays.stream(valuesArray).map(BigDecimal::new).toArray(BigDecimal[]::new);
+            if (!seriesList.stream().anyMatch(series -> hasValues(series, decimals))) {
```
unrealwork: could be replaced with `noneMatch`
```java
if (seriesList.stream().noneMatch(series -> hasValues(series, decimals)))
```

src/main/java/com/axibase/tsd/api/model/series/Series.java
```diff
@@ -16,7 +17,7 @@
 @Accessors(chain = true)
 @JsonInclude(JsonInclude.Include.NON_NULL)
 @JsonIgnoreProperties(ignoreUnknown = true)
-public class Series {
+public class Series implements Comparable<Series> {
```
unrealwork: Is it default Series comparator? May be, implement separate comparator and use its instance as argument of `sort` method?

src/test/java/com/axibase/tsd/api/method/series/PreEvaluationTest.java
```diff
@@ -0,0 +1,289 @@
...
+
+    /**
+     * Generate samples all with the same value, and following timestamps:
+     * 2019-07-01T00:00:00Z
+     * 2019-07-01T00:05:00Z
+     * 2019-07-01T00:09:00Z
+     * 2019-07-01T00:15:00Z
+     * 2019-07-01T00:19:00Z
+     */
+    private List<Sample> generateSamples(int value) {
```
unrealwork: Replace with `Collections#asList` to prevent duplication.

```java
return asList(
                Sample.ofDateDecimal("2019-07-01T00:00:00Z", BigDecimal.valueOf(value)),
                Sample.ofDateDecimal("2019-07-01T00:09:00Z", BigDecimal.valueOf(value)),
                Sample.ofDateDecimal("2019-07-01T00:05:00Z", BigDecimal.valueOf(value)),
                Sample.ofDateDecimal("2019-07-01T00:15:00Z", BigDecimal.valueOf(value)),
                Sample.ofDateDecimal("2019-07-01T00:19:00Z", BigDecimal.valueOf(value))
        );
```

src/test/java/com/axibase/tsd/api/util/CommonAssertions.java
raipc: `iff` means `if and only if`

MikhailZvagelsky: It means "if and only if" - https://en.wikipedia.org/wiki/If_and_only_if

src/test/java/com/axibase/tsd/api/method/series/PostEvaluationTest.java
MikhailZvagelsky: Changed in commit a788e1b

src/test/java/com/axibase/tsd/api/util/CommonAssertions.java
MikhailZvagelsky: That is good remark. Applied in commit a788e1b

src/test/java/com/axibase/tsd/api/util/CommonAssertions.java
MikhailZvagelsky: Changed in commit a788e1b
(But difference between `!allMatch()` and `noneMatch()` is so miserable :).)

src/main/java/com/axibase/tsd/api/model/series/Series.java
MikhailZvagelsky: Yes, that is default, so I doesn't change the code.
I think we will implement comparator when we will need another order of series.

src/test/java/com/axibase/tsd/api/method/series/PreEvaluationTest.java
MikhailZvagelsky: Replaced in commit 81bf1bb

PR #561, opened 2019-07-12
Disable testAppendWithErase for TCP transport

raipc:


src/test/java/com/axibase/tsd/api/method/series/command/AppendFieldTest.java
```diff
@@ -100,8 +102,8 @@ public void testAppendWithErase() throws Exception {
             commandList.add(seriesCommand);
         }
 
-        transport.send(new ListCommand(commandList));
-
+//        transport.send(new ListCommand(commandList));
+        Transport.HTTP.send(new ListCommand(commandList)); // Fix when TCP errors details are ready
```
Arthasasyan: Wouldn't it be better to mark this line with TODO comment? In this case it won't get lost when TCP errors details are ready.

PR #579, opened 2019-08-05
SQL insert and update tests

Arthasasyan:
Some points about this PR:
1. `INSERT` and `UPDATE` actually do the same thing in our SQL dialect. That's why I decided not to  split tests with these clauses into different classes. Except of this, new enum `InsertionType` with method that prepares SQL query was created.
2. `InsertionType.insertionQuery` could not be applied to [one method](https://github.com/axibase/atsd-api-test/blob/bb031bdf12f17b2e22caa89db4802c5789c8d777/src/test/java/com/axibase/tsd/api/method/sql/clause/insert/SqlInsertionErrorsTest.java#L28). That's because `ImmutableMap` cannot accept `null` parameters and finding another ways to use `insertionQuery` method was, IMHO, a little senseless for just one test method.
3. `INSERT INTO` and `UPDATE` raise different error messages at [this case](https://github.com/axibase/atsd-api-test/blob/bb031bdf12f17b2e22caa89db4802c5789c8d777/src/test/java/com/axibase/tsd/api/method/sql/clause/insert/SqlInsertionErrorsTest.java#L31).
4. New clauses return error messages in format different from our classical format. Please, see Issue 5962#note69 to find format differences. That's why `extractSqlErrorMessage` method was changed.


src/main/java/com/axibase/tsd/api/method/entity/EntityMethod.java
```diff
@@ -36,6 +36,10 @@ public static Response createOrReplaceEntity(Entity entity) {
         return createOrReplaceEntity(entity.getName(), entity);
     }
 
+    public static void createOrReplaceEntityCheck(String entityName) throws Exception {
```
raipc: why?

src/test/java/com/axibase/tsd/api/method/sql/clause/insert/SqlInsertionTest.java
```diff
@@ -0,0 +1,242 @@
...
+import org.testng.annotations.Test;
+
+import java.util.Collections;
+
+public class SqlInsertionTest extends SqlTest {
+    private static final String ISO_TIME = Mocks.ISO_TIME;
+    private static final long MILLIS_TIME = Mocks.MILLS_TIME;
+    private static final int VALUE = Mocks.INT_VALUE;
+
+    private final String entity1 = Mocks.entity();
```
raipc: unstable test. Don't rely on instance variables. Why not move them to the tests where they are used?

src/test/java/com/axibase/tsd/api/method/sql/clause/insert/InsertionType.java
```diff
@@ -0,0 +1,60 @@
...
+import com.axibase.tsd.api.util.TestUtil;
+import org.testng.annotations.DataProvider;
+
+import java.util.Map;
+
+public enum InsertionType {
+    INSERT_INTO {
+        @Override
+        public String insertionQuery(String tableName, Map<String, Object> columns) {
+            String columnNames = "(";
```
raipc:
```java
 String columnNames = String.join(", ", columns.keySet());
 String columnValues = columns.values().stream()
                    .map(this::encloseInQuotes)
                    .collect(Collectors.joining(", "));
 return String.format("INSERT INTO \"%s\" (%s) VALUES (%s)", tableName, columnNames, columnValues);
```

src/test/java/com/axibase/tsd/api/method/sql/clause/insert/InsertionType.java
```diff
@@ -0,0 +1,60 @@
...
+            }
+            columnNames += ")";
+            columnValues += ")";
+            return String.format("INSERT INTO \"%s\"%s VALUES%s", tableName, columnNames, columnValues);
+        }
+    },
+    UPDATE {
+        @Override
+        public String insertionQuery(String tableName, Map<String, Object> columns) {
+            String keysAndValues = "";
```
raipc:
```java
 String keysAndValues = columns.entrySet()
                    .stream()
                    .map(e -> e.getKey() + "=" + encloseInQuotes(e.getValue()))
                    .collect(Collectors.joining(", "));
  return String.format("UPDATE \"%s\" SET %s", tableName, keysAndValues);
```

src/test/java/com/axibase/tsd/api/method/sql/clause/insert/InsertionType.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public abstract String insertionQuery(String tableName, Map<String, Object> columns);
+
+
+    protected String encloseInQuotes(Object object) {
+        if (object.getClass().getSuperclass().getSimpleName().equals("Number") || object.getClass().getSimpleName().equals("ScientificNotationNumber")) {
+            return object.toString();
+        } else {
+            return String.format("'%s'", object.toString());
```
raipc: Don't call explicit toString on format args

src/main/java/com/axibase/tsd/api/method/metric/MetricMethod.java
```diff
@@ -129,6 +129,10 @@ public static void createOrReplaceMetricCheck(Metric metric) throws Exception {
         createOrReplaceMetricCheck(metric, new MetricCheck(metric));
     }
 
+    public static void createOrReplaceMetricCheck(String metricName) throws Exception {
+        createOrReplaceMetricCheck(new Metric(metricName));
```
raipc: ?

Arthasasyan: I believe, that using this method would be much more comfortable than `createOrReplaceMetricCheck(Metric)` when we meed just an empty metric. The same for `EntityMethod`

src/test/java/com/axibase/tsd/api/util/TestUtil.java
```diff
@@ -275,4 +277,18 @@ public static String quoteEscape(String s) {
...
         return mapTags;
     }
+
+    public static Map<String, Object> toUnmodifiableMap(Object... objects) {
+        if(objects.length %2 != 0) {
+            throw new IllegalArgumentException("Objects count must be even!");
+        }
+        Map<String, Object> map = new LinkedHashMap<>();
+        for(int i = 0; i < objects.length; i+=2) {
+            if(!objects[i].getClass().getSimpleName().equals("String")) {
```
raipc: instanceof

src/test/java/com/axibase/tsd/api/util/TestUtil.java
```diff
@@ -275,4 +277,18 @@ public static String quoteEscape(String s) {
 
         return mapTags;
     }
+
+    public static Map<String, Object> toUnmodifiableMap(Object... objects) {
```
raipc: Add a comment when this method should be preferred to guava

src/test/java/com/axibase/tsd/api/method/sql/clause/insertandupdate/SqlInsertIntoErrorsTest.java
```diff
@@ -0,0 +1,121 @@
...
+    @Test(
+            description = "Tests that if one of declared parameters' value is empty, error is thrown"
+    )
+    @Issue("5962")
+    public void testEmptyDeclaredParameterInsertion(){
+        String sqlQuery = INSERTION_TYPE.insertionQuery(Mocks.metric(), TestUtil.toUnmodifiableMap(
+                "entity", Mocks.entity(), "datetime", ISO_TIME, "value", VALUE, "empty", null
+                ));
+        String errorMessage = "net.sf.jsqlparser.parser.ParseException: Encountered unexpected token: \")\" \")\"\n" +
+                "    at line 1, column 359.\n" +
```
raipc: Check substring

src/test/java/com/axibase/tsd/api/util/ScientificNotationNumber.java
```diff
@@ -0,0 +1,16 @@
+package com.axibase.tsd.api.util;
+
+import java.text.DecimalFormat;
+
+public class ScientificNotationNumber {
+    private final String scientificNotation;
+
+    public ScientificNotationNumber(double number) {
```
raipc:
```java
@RequiredArgsConstructor
public class ScientificNotationNumber extends Number {
    @Delegate
    private final Number delegate;

    @Override
    public String toString() {
        return new DecimalFormat("0.0E0").format(delegate);
    }
}
```

src/test/java/com/axibase/tsd/api/method/sql/clause/insertandupdate/InsertionType.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public abstract String insertionQuery(String tableName, Map<String, Object> columns);
+
+
+    protected String encloseInQuotes(Object object) {
+        if(object == null) {
+            return "";
+        }
+        if (object instanceof Number || object instanceof ScientificNotationNumber) {
```
raipc: instanceof Number

src/test/java/com/axibase/tsd/api/method/sql/clause/insertandupdate/SqlInsertNonExistentEntityTest.java
```diff
@@ -0,0 +1,89 @@
...
+    @Test(
+            description = "Tests that entity is created after INSERT INTO command if it did not exist."
+    )
+    @Issue("5962")
+    public void testEntityCreationWhenInserting() {
+        String entity = Mocks.entity();
+        String sqlQuery = insertionType.insertionQuery(metric,
+                ImmutableMap.of("entity", entity, "datetime", ISO_TIME, "value", VALUE));
+        assertOkRequest("Insertion of series with nonexistent entity failed!", sqlQuery);
+        Checker.check(new EntityCheck(
```
raipc: Initialize Entity object using constructor prior to submitting the query to assure that entity did not exist before query execution. Same for all tests that check metric or entity insertion

src/test/java/com/axibase/tsd/api/method/sql/clause/insertandupdate/SqlUpdateErrorsTest.java
```diff
@@ -0,0 +1,110 @@
...
+import com.axibase.tsd.api.util.TestUtil;
+import com.google.common.collect.ImmutableMap;
+import io.qameta.allure.Issue;
+import org.testng.annotations.Test;
+
+public class SqlUpdateErrorsTest extends SqlTest {
+    private static final String ISO_TIME = Mocks.ISO_TIME;
+    private static final int VALUE = Mocks.INT_VALUE;
+
+    private static final InsertionType INSERTION_TYPE = InsertionType.UPDATE;
```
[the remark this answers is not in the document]
Arthasasyan: Could you explain why it should be inlined? I just don't see the point because this constant is used in all methods of the class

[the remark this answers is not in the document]
raipc: I read the test method, find that variable usage and think: "Wait, why is INSERTION_TYPE is used here if the class name is called SqlUpdateTest?" Then I need to go to the top to see the actual value. If I see `InsertionType.UPDATE` right in the method it would be easier to understand.

PR #582, opened 2019-09-02
Added tests for DATEADD with literal dates

Arthasasyan:
Pending for problems with add moths and add quarters to be resolved

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
@@ -43,7 +43,7 @@ public class DateAddTest extends SqlTest {
                 result.add(new Object[]{datetime[0], unitAndResult[0], unitAndResult[1]});
             }
         }
-        return result.toArray(new Object[dateTime().length * 3][3]);
+        return result.toArray(new Object[dateTime().length * timeUnitAndExpectedResult().length][3]);
     }
 
     @Test(
--- a/src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
@@ -27,7 +27,6 @@ public class DateAddTest extends SqlTest {
     }
 
     private String[][] timeUnitAndExpectedResult() {
-        ;
         return new String[][] {
                 {"second", "2019-01-01 00:00:01"},
                 {"minute", "2019-01-01 00:01:00"},
--- a/src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
@@ -2,6 +2,7 @@ package com.axibase.tsd.api.method.sql.function.date;
 
 import com.axibase.tsd.api.method.sql.SqlTest;
 import com.axibase.tsd.api.util.TestUtil;
+import com.axibase.tsd.api.util.Util;
 import io.qameta.allure.Issue;
 import org.testng.annotations.DataProvider;
 import org.testng.annotations.Test;
@@ -9,11 +10,12 @@ import org.testng.annotations.Test;
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
-import java.util.TimeZone;
 
 public class DateAddTest extends SqlTest {
     private static final String SQL_QUERY_TEMPLATE = "SELECT date_format(DATEADD(%s, 1, '%s'), 'yyyy-MM-dd HH:mm:ss')";
-    private static final Date DATE = new Date(1546300800000L); //2019-01-01 00:00:00
+    private static final Date DATE = new Date(Util.parseAsServerZoned("2019-01-01T00:00:00")
+            .toInstant()
+            .toEpochMilli());
 
     private String[] patterns() {
         return new String[]{
@@ -22,7 +24,7 @@ public class DateAddTest extends SqlTest {
                 "yyyy-MM-dd",
                 "yyyy-MM-dd HH:mm:ss",
                 "yyyy-MM-dd HH:mm:ss.SSS",
-                TestUtil.isoPatternWithOffset(DATE)
+                "yyyy-MM-dd'T'HH:mm:ss.SSSXXX"
         };
     }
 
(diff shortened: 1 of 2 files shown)


Arthasasyan:
The problem was in the wrong pattern

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -0,0 +1,126 @@
...
+                {"2019-01-01T00:00:00.000+05:45"} //Kathmandu timezone
+        };
+    }
+
+    @Test(
+            dataProvider = "dateTime",
+            description = "Test DATEADD with second argument"
+    )
+    @Issue("6528")
+    public void testAddSecond(String isoTime) {
```
raipc: Call the argument `datetime`. Not all values returned by data provider are in ISO format.

raipc: interval unit can be also parameterized

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -0,0 +1,126 @@
+package com.axibase.tsd.api.method.sql.function.date;
+
+import com.axibase.tsd.api.method.sql.SqlTest;
+import io.qameta.allure.Issue;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+public class DateAddTest extends SqlTest {
+    private static final String SQL_QUERY_TEMPLATE = "SELECT date_format(DATEADD(%s, 1, '%s'), 'YYYY-MM-dd HH:mm:ss')";
```
raipc: Always use yyyy instead of YYYY. [Week-based year](https://axibase.com/docs/atsd/shared/time-pattern.html#date-and-time-patterns) should be used only in very special cases.

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
Arthasasyan: I can use dataprovider like this:

```java
@DataProvider
    public Object[][] calendarUnit() {
        return new Object[][] {
                {"second", Calendar.SECOND},
                //...
        };
    }
```

The problem is that `Calendar` enum has no constant for quarter and it seems that I cannot use field like `{"quarter", 3 * Calendar.Month}`

raipc: You don't need an extra data provider in this case. Refactor an existing one to a data provider with fields dateTime, timeUnit and expectedResult. It's not necessary to use constants for time units, but there is an enum TimeUnit in the project which defines supported units including quarter.

Arthasasyan: Changed with the latest commit. I forced dataprovider to collect data itself, because it's much easier then printing and manually managing 48 combinations

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -43,7 +43,7 @@
                 result.add(new Object[]{datetime[0], unitAndResult[0], unitAndResult[1]});
             }
         }
-        return result.toArray(new Object[dateTime().length * 3][3]);
+        return result.toArray(new Object[dateTime().length * timeUnitAndExpectedResult().length][3]);
```
raipc: You don't need to calculate array size, just pass `new Object[0][0]`

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -0,0 +1,62 @@
...
+    private static final String SQL_QUERY_TEMPLATE = "SELECT date_format(DATEADD(%s, 1, '%s'), 'yyyy-MM-dd HH:mm:ss')";
+
+    private Object[][] dateTime() {
+        return new String[][] {
+                {"2019"},
+                {"2019-01"},
+                {"2019-01-01"},
+                {"2019-01-01 00:00:00"},
+                {"2019-01-01 00:00:00.000"},
+                {"2019-01-01T00:00:00.000+05:45"} //Kathmandu timezone
```
raipc: Last value is not equal to previous. It is 31 Dec in UTC.

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -0,0 +1,62 @@
...
+import com.axibase.tsd.api.method.sql.SqlTest;
+import io.qameta.allure.Issue;
+import org.testng.annotations.DataProvider;
+import org.testng.annotations.Test;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public class DateAddTest extends SqlTest {
+    private static final String SQL_QUERY_TEMPLATE = "SELECT date_format(DATEADD(%s, 1, '%s'), 'yyyy-MM-dd HH:mm:ss')";
```
raipc: It seems that the test will pass only if server timezone is Kathmandu

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -0,0 +1,62 @@
...
+
+    private Object[][] dateTime() {
```
raipc: 1-dim array is enough

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
raipc: It would be better to create the datetime array by specifying patterns and formatting a single timestamp in a non-default time zone. [date-processor](https://github.com/axibase/date-processor) library should do it fast. It would be nice to use it instead of multiple libraries in Util class (but Util methods modification should be done in a separate PR)

Arthasasyan:
Pending for #586

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -1,28 +1,34 @@
...
-                {"2019-01-01 00:00:00.000"},
-                {"2019-01-01T00:00:00.000+05:45"} //Kathmandu timezone
+    private String[] patterns() {
+        return new String[]{
+                "yyyy",
+                "yyyy-MM",
+                "yyyy-MM-dd",
+                "yyyy-MM-dd HH:mm:ss",
+                "yyyy-MM-dd HH:mm:ss.SSS",
+                String.format("yyyy-MM-dd'T'HH:mm:ss.SSS'%tz'", Instant.ofEpochMilli(DATE.getTime()).atZone(Util.getServerTimeZone().toZoneId()))
```
raipc: Format using TestUtil.formatDate

src/test/java/com/axibase/tsd/api/method/sql/function/date/DateAddTest.java
```diff
@@ -59,4 +66,9 @@ public void dateAddTest(String datetime, String timeUnit, String expectedResult)
         assertSqlQueryRows(sqlQueryRow, sqlQuery);
     }
 
+    private String simpleDateFormat(String pattern) {
```
raipc: Use TestUtil instead of copying date formatter creation from test to test.

PR #597, opened 2019-12-06
Fixed error messages with tags

Arthasasyan:


Arthasasyan:
Continue Redmine issue 6713

PR #603, opened 2020-03-12
Fix TokenMessageTest.testStatsMethod.

MikhailZvagelsky:
Seems that previous implementation of the message stats service manually sets queried alignment to START_TIME. Current implementation uses default calendar alignment, and the test fails. So I set START_TIME alignment in the test to fix this discrepancy.


raipc:
@MikhailZvagelsky Do you know what 3 time zones are used in Jenkins builds?

PR #604, opened 2020-05-07
Add test for financial mode, instrument search endpoint

VeselovAlex:


VeselovAlex:
Blocked until index refresh endpoint will not be merged into master

src/test/java/com/axibase/tsd/api/method/financial/InstrumentSearchBase.java
```diff
@@ -0,0 +1,175 @@
...
+        public static long timestamp() {
+            return timestamp.addAndGet(5);
+        }
+
+        public static long trade() {
+            return trade.incrementAndGet();
+        }
+    }
+
+    private static class InstrumentSearchResponse extends GenericType<List<InstrumentSearchEntry>> {
```
raipc, line 172: Does this approach work? As I remember, Jersey needs to know list type parameter which gets erased when generic class is inherited, but it is kept when the child class inherits type parameter, e.g. in anonymous classes. See examples in `ResponseAsList` class

rodionos:
@VeselovAlex - Who is blocking this PR?

VeselovAlex:
Waiting for tsd#220

rodionos:
@VeselovAlex  Can you reference my nickname in reply so that I get notifications?

raipc:
Good to merge when travis build passes

rodionos:
@unrealwork Do you mean Registry which we use for new entities, but apply it to instrument names?

unrealwork:
> @unrealwork Do you mean Registry which we use for new entities, but apply it to instrument names?

@rodionos Yes, Trades CSV API creates new entities that are not under control of our Registry.
And every atsd named element (metric, entity, replacemen-table, etc) better to name via [Registry](https://github.com/axibase/atsd-api-test/blob/master/src/main/java/com/axibase/tsd/api/util/Registry.java).

PR #610, opened 2020-05-21
SQL test for invalid instrument conditions

alexandertokarev:
API tests for axibase/tsd#254

### DIFF
--- a/src/test/java/com/axibase/tsd/api/method/sql/trade/TradeInstrumentInvalidConditionTest.java
+++ b/src/test/java/com/axibase/tsd/api/method/sql/trade/TradeInstrumentInvalidConditionTest.java
@@ -1,6 +1,8 @@
 package com.axibase.tsd.api.method.sql.trade;
 
+import com.axibase.tsd.api.method.entitygroup.EntityGroupMethod;
 import com.axibase.tsd.api.method.sql.SqlTest;
+import com.axibase.tsd.api.model.entitygroup.EntityGroup;
 import com.axibase.tsd.api.model.financial.Trade;
 import com.axibase.tsd.api.util.TradeSender;
 import org.testng.annotations.BeforeClass;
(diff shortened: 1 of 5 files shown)


src/test/java/com/axibase/tsd/api/method/sql/trade/TradeInstrumentInvalidConditionTest.java
```diff
@@ -0,0 +1,121 @@
...
+        String sqlQuery = "select * from atsd_trade where symbol like '" + TEST_PREFIX + "symbol1%' and symbol in ('" + TEST_PREFIX + "symbol3')";
+        assertBadSqlRequest("Mutual exclusive conditions: symbol like '" + TEST_PREFIX + "symbol1%', symbol in ('" + TEST_PREFIX + "symbol3')", sqlQuery);
+
+        sqlQuery = "select * from atsd_trade where class like '" + TEST_PREFIX + "class1%' and class = '" + TEST_PREFIX + "class3'";
+        assertBadSqlRequest("Mutual exclusive conditions: class like '" + TEST_PREFIX + "class1%', class = '" + TEST_PREFIX + "class3'", sqlQuery);
+    }
+
+    @Test
+    public void testMutualExclusiveConditionEntityGroup() throws Exception {
+        EntityGroupMethod.createOrReplaceEntityGroup(new EntityGroup(TEST_ENTITY_GROUP2_NAME));
```
raipc, line 111: Could you please move entity group setup to BeforeClass method and create a data provider that returns SQL query, expected error, and query description, and use this provider in a single method with assertBadSqlRequest statement with 3 arguments? In this case all queries will be tested, not only until first failure? Also it will be easier to check that different queries operate with different instruments or entity groups

raipc:
Tests pass after merging axibase/tsd#254

PR #698, opened 2021-03-17
Support multiple values for a tag in series query.

MikhailZvagelsky:
Change `SeriesQuery.tags` type from `Map<String, String>` to `Map<String, List<String>>` to allow multiple values for a tag in API series query.

### DIFF
--- a/src/main/java/com/axibase/tsd/api/util/Util.java
+++ b/src/main/java/com/axibase/tsd/api/util/Util.java
@@ -106,14 +106,7 @@ public class Util {
         return map.entrySet().stream().collect(Collectors.toMap(x -> x.getKey(), x -> x.getValue()));
     }
 
-    /** Wrap each value of provided map into list. */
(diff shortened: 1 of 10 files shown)


raipc (approved):

src/main/java/com/axibase/tsd/api/util/Util.java
```diff
@@ -109,6 +106,19 @@ public static String prettyPrint(Object o) {
...
     }
 
+    /** Wrap each value of provided map into list. */
+    public static Map<String, List<String>> wrapValuesInLists(Map<String, String> map) {
+        return map.entrySet().stream().collect(
+                Collectors.toMap(Map.Entry::getKey, entry -> wrapInList(entry.getValue()))
+        );
+    }
+
+    public static ArrayList<String> wrapInList(String value) {
```
raipc, line 116: I think the name should reflect that returned list is mutable, otherwise not clear why Arrays.asList not used

PR #750, opened 2021-07-26
Don't throw registry error on test method retry

raipc:
Related to [retry issue](https://github.com/axibase/tsd/issues/1785#issuecomment-884679616)


raipc:
The change eliminates the requirement to insert test data in static blocks, so the exception during it (e.g. Jetty header parse error) doesn't prevent from executing test methods again

PR #757, opened 2021-08-09
Tests SQL daily time range conditions

alexandertokarev:


MikhailZvagelsky:
Sasha, that is because my `CurrentDayBarsTest` fails. I am considering it.

PR #818, opened 2022-10-31
Run tests with testcontainers

raipc:


src/main/java/com/axibase/tsd/api/util/AtsdContainer.java
```diff
@@ -63,26 +64,28 @@ private HttpWaitStrategy waitUntilUserAppears() {
...
         atsdContainer.loadWorkdayCalendars();
         return atsdContainer;
     }
 
     private void loadWorkdayCalendars() {
-        try (final CloseableHttpClient httpClient = HttpClients.createMinimal()) {
-            httpClient.execute(prepareCalendarUpload("rus"), new BasicResponseHandler());
-            httpClient.execute(prepareCalendarUpload("usa"), new BasicResponseHandler());
+        try (final CloseableHttpClient httpClient = HttpClients.createSystem()) {
+            httpClient.execute(prepareCalendarUpload("rus"), new BasicHttpContext(), new BasicHttpClientResponseHandler());
```
unrealwork, line 73: Should list of calendars be configurable?

raipc: No, I think it should be hardcoded. There are many configuration options that we don't use, and I intend to reduce them

